PR #10, opened 2016-09-13
feat(rsa): always use the fastest method available for rsa keygeneration

### LINKED ISSUE #9: Enable WebWorkers to make it go faster
node-forge supports WebWorkers
image

enable here: https://github.com/libp2p/js-libp2p-crypto/blob/master/src/keys/rsa.js#L131

dignifiedquire:
BREAKING CHANGE: generateKeyPair is now async

Simple secio benchmarks of key gen, (node.js)
- before this change: 0.27 ops/sec
- after this change (with ursa): 8.47 ops/sec
- after this change (without ursa): 0.52 ops/sec

Fixes #9

### DIFF
--- a/package.json
+++ b/package.json
@@ -2,7 +2,7 @@
   "name": "libp2p-crypto",
   "version": "0.6.1",
   "description": "Crypto primitives for libp2p",
-  "main": "lib/index.js",
+  "main": "src/index.js",
   "jsnext:main": "src/index.js",
   "scripts": {
     "lint": "aegir-lint",
@@ -55,5 +55,8 @@
     "Friedel Ziegelmayer <dignifiedquire@gmail.com>",
     "Richard Littauer <richard.littauer@gmail.com>",
     "greenkeeperio-bot <support@greenkeeper.io>"
-  ]
+  ],
+  "optionalDependencies": {
+    "ursa": "^0.9.4"
+  }
 }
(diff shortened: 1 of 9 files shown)


package.json
```diff
@@ -2,7 +2,7 @@
   "name": "libp2p-crypto",
   "version": "0.6.1",
   "description": "Crypto primitives for libp2p",
-  "main": "lib/index.js",
+  "main": "src/index.js",
```
dignifiedquire: Todo: revert before merge

src/keys/rsa.js
```diff
@@ -127,9 +127,39 @@ function unmarshalRsaPublicKey (bytes) {
...
 
-function generateKeyPair (bits) {
-  const p = rsa.generateKeyPair({bits})
-  return new RsaPrivateKey(p.privateKey, p.publicKey)
+function fastKeys (bits, cb) {
+  if (typeof window === 'undefined') {
+    let ursa
+    try {
+      ursa = require('ursa')
+    } catch (err) {
```
VictorBjelkholm: Should at least show a warning if something goes wrong here

dignifiedquire: Why? This is simply the check if the optional dependency is available to be used

VictorBjelkholm: Ah, did not know is what an optional dependency and my spidey senses goes off when I see a empty catch in a try/catch block.

dignifiedquire: As they should :)

dignifiedquire:
Closing in favor of #12

PR #16, opened 2016-10-19
package.json - fix main reference

kumavis:


kumavis:
see https://github.com/libp2p/js-libp2p-floodsub/pull/6

PR #12, opened 2016-09-23
Async Crypto Endeavour

dignifiedquire:
- No more `node-forge`
- No more `elliptic.js`

~~Needs https://github.com/multiformats/js-multihashing/pull/12~~ Now `multihashing-async`

Stats can be found here: https://github.com/libp2p/js-libp2p-crypto/blob/webcrypto/stats.md

PR #17, opened 2016-10-19
browser fix - pre-load files with brfs

kumavis:
We are loading a file via `fs` here https://github.com/libp2p/js-libp2p-crypto/blob/master/src/index.js#L6

This breaks support for module consumers using [browserify](https://github.com/substack/node-browserify). The [brfs](https://github.com/substack/brfs) module fixes this. The flag in the `package.json` indicates that this module requires pre-processing via the local brfs dependency to be imported correctly.

### DIFF
--- a/package.json
+++ b/package.json
@@ -28,7 +28,8 @@
     "elliptic": "^6.3.2",
     "multihashing": "^0.2.1",
     "node-forge": "^0.6.39",
-    "protocol-buffers": "^3.1.6"
+    "protocol-buffers": "^3.1.6",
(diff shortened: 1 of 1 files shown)


kumavis:
this fix is also required for browserify support: https://github.com/libp2p/js-libp2p-crypto/pull/16

dignifiedquire:
#12 removes the use of `brfs` entirely and is hopefully getting shipped in the next two weeks.

kumavis:
I'll wait 👍

PR #29, opened 2016-11-10
AES Interop

dignifiedquire:
**Challenge:** Ensure three implementations of `AES CTR 128`  and `AES CTR 256` work the same.

**General Spec:** [NIST SP 800-38A](http://nvlpubs.nist.gov/nistpubs/Legacy/SP/nistspecialpublication800-38A.pdf)

#### 1. SubtleCrypto
- Spec: https://www.w3.org/TR/WebCryptoAPI/#aes-ctr-description
- Actual implmentations in Chromium and Firefox
- Usage: https://github.com/libp2p/js-libp2p-crypto/blob/master/src/crypto/aes-browser.js

#### 2. Node.js `crypto` module
- Actually uses OpenSSL
- Usage: https://github.com/libp2p/js-libp2p-crypto/blob/master/src/crypto/aes.js

#### 3. golang
- https://golang.org/src/crypto/aes/cipher.go
- https://golang.org/src/crypto/cipher/ctr.go
- Usage:
  - Init: https://github.com/libp2p/go-libp2p-secio/blob/master/al.go
  - Encrypt: https://github.com/libp2p/go-libp2p-secio/blob/master/rw.go#L57
  - Decrypt: https://github.com/libp2p/go-libp2p-secio/blob/master/rw.go#L237

### Generating Fixtures

- Samples Program to ensure compat with go from the JavaScript side: https://gist.github.com/dignifiedquire/b8aee1b2b8a8903d2f96b657ed0cad6e (Results are in `test/fixtures/go-aes.js`
- Write node.js generated data to `test/fixtures/aes.js` 

### Current State

- Node.js & go generate the same results, after multiple encryptions
- webcrypto generates a different result after the first round, indicating the counter increments and/or reuse are different.


dignifiedquire:
fix(aes): replace subtle.crypto with browserify-aes  …
Due to the design of `AES-CTR` in the webcrypto spec, there
is no streaming mode provided. This results in the counter
not being reused between subsequent calls to `encrypt` or
`decrypt`. As both the node.js and the go implementation rely
on this webcrypto had to be replaced.

dignifiedquire:
@diasdavid this is passing and ready for review and merge

diasdavid:
So, there is no streaming mode in AES-CTR? Then it is just AES-ECB. Having it being a stream, something you can add more that to encrypt making every encrypted block dependent on the previous, is one of the things that gives strength to AES encryption, that is why AES-ECB is not used at all anymore (or should not). 

This must be a bug in WebCrypto (or somewhere along the way), have you seen this: https://github.com/diafygi/webcrypto-examples#aes-ctr---encrypt?

Thank you for coming up with an alternative solution though :)

PR #42, opened 2016-11-29
Better compatability

### LINKED ISSUE #38: Find alternative to node-webcrypto-ossl
Sadly we will have to let go of node-webcrypto-ossl as it is blocking users from using it in electron.

Details can be found here:

- https://github.com/PeculiarVentures/node-webcrypto-ossl/issues/77
- https://github.com/electron/electron/issues/1410#issuecomment-152407010

We currently use

- ECDH: https://github.com/libp2p/js-libp2p-crypto/blob/master/src/crypto/ecdh.js
- RSASSA-PKCS1-v1_5: https://github.com/libp2p/js-libp2p-crypto/blob/master/src/crypto/rsa.js

Of which 
- ECDH is available in Node.js core: https://nodejs.org/docs/latest/api/crypto.html#crypto_class_ecdh but we have to check if this provides everything we need. 
- RSA support is not complete in Node.js core as far as I understand, so we will need a different module for this.

dignifiedquire:
Plan

- [x] feat(ecdh): use node core instead of webcrypto-ossl
- [x] feat(rsa): fall back to pure js version if webcrypto-ossl is not installed

Fixes #38


dignifiedquire:
@diasdavid @haadcode please check this out :) no more reliance on native dependencies, but they are used if available to make things faster.

diasdavid:
will review tonight or first thing tomorrow morning. Thank you!

haadcode:
LGTM in general. Have you tested this with js-ipfs-api to see that this indeed solves the problem of npm installing it and using with electron?

PR #45, opened 2016-12-01
feat: add randomBytes function

diasdavid:
adds feature needed for https://github.com/libp2p/js-libp2p-ping/pull/38

> PS: I got very confused by the amount of times that the words `crypto` and `webcrypto` appear interleave throughout the code (ie. crypto = require(webcrypto) & crypto.webcrypto 💥).

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -71,3 +71,12 @@ exports.marshalPrivateKey = (key, type) => {
 
   return key.bytes
 }
+
+exports.randomBytes = (number) => {
+  if (!number || typeof number !== 'number') {
+    throw new Error('first argument must be a Number')
+  }
+  const buf = new Buffer(number)
+  c.webcrypto.getRandomValues(buf)
+  return buf
+}
--- a/src/index.js
+++ b/src/index.js
@@ -74,9 +74,9 @@ exports.marshalPrivateKey = (key, type) => {
 
 exports.randomBytes = (number) => {
   if (!number || typeof number !== 'number') {
-    throw new Error('first argument must be a Number')
+    throw new Error('first argument must be a Number bigger than 0')
   }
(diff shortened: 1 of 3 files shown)


src/index.js
```diff
@@ -71,3 +71,12 @@ exports.marshalPrivateKey = (key, type) => {
 
   return key.bytes
 }
+
+exports.randomBytes = (number) => {
```
dignifiedquire: This already exists: 
- node: https://github.com/libp2p/js-libp2p-crypto/blob/master/src/crypto/rsa.js#L58
- browser: https://github.com/libp2p/js-libp2p-crypto/blob/master/src/crypto/rsa-browser.js#L52

you just need to expose it

diasdavid: 11:39 <@daviddias> that " needs to be exposed" is difference
11:39 <@daviddias> because what I really wanted is to give a length and get a buffer
11:39 <@daviddias> I updated my PR to use that function
11:39 <@daviddias> but it is just adding a op

src/index.js
```diff
@@ -71,3 +71,12 @@ exports.marshalPrivateKey = (key, type) => {
...
+  if (!number || typeof number !== 'number') {
```
dignifiedquire: technically this will throw on `0` which is a number ;)

dignifiedquire:
LGTM

src/index.js
diasdavid: which is ok, as there is no `0 random bytes`

diasdavid: added a more clear error message to say it has to be > 0

Kubuxu: You are not checking if it positive.
<sup>but it might be not needed. /me has no idea how JS behaves</sup>

Issue #41, opened 2016-11-28
Uncaught error when using browserify


JustinDrake:
Below is my code. I use Browserify and run the code in Chrome. The code runs `crypto.generateKeyPair` without issues, but chokes on `peerId.createFromPrivKey` with an Uncaught Exception. See screenshot.

```js
const crypto = require('libp2p-crypto')
const peerId = require('peer-id')

// Example: create an identity
crypto.generateKeyPair('RSA', 2048, (err, key) => {
	if(err) {
		console.error(err);
		return err;
	}

	// So far so good!
	console.log(key);

	// Throws an Uncaught Error!
	peerId.createFromPrivKey(key.bytes, (err, id) => {
		if(err) {
			console.error(err);
			return err;
		}

		console.log(id.toB58String());
	});
});
```

<img width="474" alt="screen shot 2016-11-23 at 12 09 30" src="https://cloud.githubusercontent.com/assets/731743/20561335/06aa31fe-b176-11e6-9760-80af8b354515.png">

JustinDrake:
It doesn't seem to work for me :(

(Still on Chrome with Browserify.)

<img width="849" alt="screen shot 2016-12-04 at 09 46 14" src="https://cloud.githubusercontent.com/assets/731743/20865325/98eb9f10-ba06-11e6-8245-6384e5e6214a.png">

dignifiedquire:
@JustinDrake it doesn't look like you are using the fixed version from the stack trace, are you using `master` from here? The fix was not released yet

JustinDrake:
I think I am using the tip of master. Below is `git log`. What should I do?

<img width="562" alt="screen shot 2016-12-04 at 09 52 47" src="https://cloud.githubusercontent.com/assets/731743/20865364/7b7261de-ba07-11e6-9d29-50ce7772ba9a.png">

JustinDrake:
My bad. For posterity, I installed the latest `master` with the command

`npm install git+https://git@github.com/libp2p/js-libp2p-crypto`

PR #47, opened 2016-12-03
Fix for browserify 

nikuda:
`util.toBase64` would throw an error as `bn.toBuffer` doesn't work with browserify. `bn.toArrayLike` is used instead. 

`bn.toBuffer` is a simple wrapper for `bn.toArrayLike` anyway so there are no downsides here between node/browser.

https://github.com/indutny/bn.js/blob/master/lib/bn.js#L516

Fixes: #41


dignifiedquire:
Sweet thank you :octocat:

PR #55, opened 2016-12-22
Add Ed25519 support for signing & verification

### LINKED ISSUE #43: Implement ed25519
Ref https://github.com/libp2p/go-libp2p-crypto/pull/5

Those will be much faster than RSA key generation :)

yusefnapora:
This implements Ed25519 support with the [TweetNaCl.js](https://github.com/dchest/tweetnacl-js) library.  

I based everything off of the RSA implementation, although of course it's missing `encrypt` and `decrypt` methods.  The go interop fixtures were generated in a [gore](https://github.com/motemen/gore) repl session using the current master branch of go-libp2p-crypto.  The `marshal` method just returns the public key bytes in a buffer for public keys, and the concatenation of secretKey + publicKey bytes for the private key, which seems to match the go implementation.

Let me know if it's missing anything or if things need tweaking.

Closes #43


src/crypto/ed25519.js
```diff
@@ -0,0 +1,33 @@
+'use strict'
+
+const nacl = require('tweetnacl')
+
+exports.publicKeyLength = nacl.sign.publicKeyLength
+exports.privateKeyLength = nacl.sign.secretKeyLength
+
+exports.generateKey = function (callback) {
+  const done = (err, res) => setImmediate(() => {
```
dignifiedquire: Please use `setImmediate = require('async/setImmediate')` to ensure we have setImmediate actually available. (yes this is not yet consistently fixed in the code base :( )

src/index.js
```diff
@@ -39,8 +41,7 @@ exports.unmarshalPublicKey = (buf) => {
 exports.marshalPublicKey = (key, type) => {
   type = (type || 'rsa').toLowerCase()
 
-  // for now only rsa is supported
-  if (type !== 'rsa') {
+  if (type !== 'rsa' && type !== 'ed25519') {
```
dignifiedquire: Now that we have more than on key type (:tada:) could you add a constant at the top `KEY_TYPES`, and check if the type is included in that?

test/fixtures/go-key-ed25519.js
```diff
@@ -0,0 +1,10 @@
+'use strict'
+
+module.exports = {
```
dignifiedquire: Could leave a comment with the commands you used to generate these? (I forgot to do that with the other ones, and missed it once bitterly)

src/index.js
yusefnapora: Sweet, just pushed this up.  I used `if (KEY_TYPES.indexOf(type) < 0)` for this, since I can't remember when `Array.includes()` became a thing or whether you're using any polyfills for this lib.  😄

dignifiedquire: Support is not great yet for `Array.prototype.includes` I think, but feel free to add `lodash.includes`which we use in other places in our modules.

dignifiedquire:
Thank you 🎉

yusefnapora:
@dignifiedquire no problem!  Glad to help out 😄

PR #56, opened 2016-12-23
fix: consistent buffer usage

### LINKED ISSUE #49: Node 4.4 and lower not supported
Discovered this while working on https://github.com/libp2p/js-libp2p-kad-routing/pull/57

On node 4.4 and below, running the the tests on this module gives a bunch of these:

```
TypeError: base64 is not a function
      at Function.from (native)
      at Function.from (native)
      at toBn (src/crypto/util.js:18:24)
      at Object.exports.jwkToPkix (src/crypto/rsa-utils.js:109:18)
      at RsaPublicKey.marshal (src/keys/rsa.js:20:25)
      at RsaPublicKey.bytes (src/keys/rsa.js:26:18)
      at Object.exports.generateKeyPair.exports.unmarshalPublicKey.exports.marshalPublicKey (src/index.js:47:13)
      at Context.<anonymous> (test/index.spec.js:67:21)
```

Which is pretty easily fixed by changing the [`toBn` function](https://github.com/libp2p/js-libp2p-crypto/blob/master/src/crypto/util.js#L18) to use the old-style Buffer constructor if `Buffer.from` fails:

```
exports.toBn = function toBn (str) {
  try {
    return new BN(Buffer.from(str, 'base64'))
  } catch (err) {
    return new BN(new Buffer(str, 'base64'))
  }
}
```

Unfortunately you can't just check for the existence of `Buffer.from`, since it does exist but is broken with the base64 argument.

But there's also a bunch of these: 

```
TypeError: this is not a typed array.
      at Function.from (native)
      at Object.module.exports.crypto.hmac.create.m.digest.m.digest [as keyStretcher] (src/key-stretcher.js:37:23)
      at Context.<anonymous> (test/key-stretcher.spec.js:37:18)
```

which I don't know how to solve 😄  

Changing the `engines` requirement in package.json to `>=4.5.0` would be a good idea until a real solution can be found.

dignifiedquire:
Closes #49


dignifiedquire:
@diasdavid please review, this is important to make sure things work in node4

PR #60, opened 2017-01-18
Implement secp256k1 with secp256k1-node module

yusefnapora:
This adds a new key type, `Secp256k1`, which does ECDSA signing / verification on the sha2-256 digest of a given message using the secp256k1 curve used in bitcoin and ethereum (and their many derivatives).  It uses the [secp256k1-node module](https://github.com/cryptocoinjs/secp256k1-node), which despite its name works just fine in the browser.  On node it will use a c implementation for a ~10x speed boost, but the pure-js implementation seems to work fine as well.

We're planning to use these keys to implement verification of ethereum signatures, but their signature scheme is very non-standard, (keccak-256 hash, with a custom signature encoding format and a fixed message prefix).  It seemed better to use sha256 + DER encoded signatures, rather than add code to libp2p that's specific to ethereum.

The public key is marshaled as a compressed 33-byte point, using the bitcoin compression scheme.  Since the public key can be derived from the private key, the private key is marshaled as just the raw 32-byte buffer.

@vyzo is working on the go side (https://github.com/libp2p/go-libp2p-crypto/issues/19) - once that's done we can add some interop tests.

I'm happy to tweak this if there's anything missing or incorrect :)


dignifiedquire:
Thank you!

Could you run `npm run build` and list the the size differences of the browser builds with and without this patch please, so I can get an understanding of how much this is adding in terms of size.

yusefnapora:
sure, no problem.  It looks like node 4 doesn't like that I used destructuring syntax also, so I'll update that too :)

yusefnapora:
I'm getting this on master:

```
Version: webpack 2.1.0-beta.28
Time: 1136ms
       Asset    Size  Chunks             Chunk Names
    index.js  563 kB       0  [emitted]  main
index.js.map  692 kB       0  [emitted]  main
[12:32:37] all files 349.78 kB

```

and on the secp256k1 branch:

```
Version: webpack 2.1.0-beta.28
Time: 1499ms
       Asset    Size  Chunks             Chunk Names
    index.js  776 kB       0  [emitted]  main
index.js.map  950 kB       0  [emitted]  main
[12:31:09] all files 499.54 kB

```

So it looks like it does add some weight to the browser build 😐 

This could probably be improved somewhat; I'll see if I can trim it down by requiring the pure-js version explicitly.  There's actually two js implementations in the module, one backed by the elliptic module, and one custom. I'm guessing they're both getting pulled in.

yusefnapora:
It looks like using the custom js implementation shaves off about 100KB from the browser build vs the default elliptic-backed implementation, and it's slightly faster (21ms for the signing test vs 24ms on my machine).  That seems like a win :)

dignifiedquire:
I am wondering if it makes sense to make these individual modules rather than pullling them in by default. Are you using this module directly or through js-ipfs?

yusefnapora:
We're using it directly.  Having separate modules sounds like a good call, since not everyone will want or need these keys.

yusefnapora:
I'm cool with making this modular, but wanted to check in first to make sure I do it in a way that makes sense for the project.  My basic plan is:

- keep the `Secp256k1` enum constant in the protobuf here in this module
- make a `js-libp2p-crypto-secp256k1` module that provides what's now in `keys/secp256k1.js` and `crypto/secp256k1.js`
- add an `optionalDependency` on `js-libp2p-crypto-secp256k1`
- in `keys/index.js`, try to require `js-libp2p-crypto-secp256k1`, and if successful, export it
- change the `KEY_TYPES` constant from hardcoded strings to the ES5 equivalent of `Object.keys(require('./keys))`
- update the unmarshal functions in `index.js` to avoid accessing `keys.secp256k1` if it doesn't exist

That seems like it should work alright; what do you think @dignifiedquire?

dignifiedquire:
Sounds good, except the `optionalDependency` part, because that means npm will still install it by default. 

Creating a repo for PR no.

dignifiedquire:
Created repo: https://github.com/libp2p/js-libp2p-crypto-secp256k1

yusefnapora:
awesome, thanks :) I didn't know that about `optionalDepenency`; that's good to know.

yusefnapora:
The last commit adds the optional support, if `require('libp2p-crypto-secp256k1')` succeeds.  I've tested it locally, but I'm not sure how to unit test it without declaring the `optionalDependency`.  

Although I guess I could at least test that it throws an exception if you try to use it without the module installed... 🤔

dignifiedquire:
You could use [rewire](https://www.npmjs.com/package/rewire) to create cases where the module exists and where it doesn't, without having to install it.

dignifiedquire:
Would also be good to add some documentation to the readme about the new module and the available functionality.

yusefnapora:
Sweet, I updated the readme and added some coverage, but I think my plan of adding the module as a `devDependency` was flawed; it's still getting rolled into the browser bundle.  should have checked that before pushing...

I'd like to test with the module present, but I guess I should just mock it out instead of depending on the real thing.  Will do that in a sec :)

yusefnapora:
There we go :) It should still be hitting all the same code paths in this module, since I'm only testing the marshal / unmarshal and generateKeyPair functions from index.js.  Everything else was just duplicating tests from the libp2p-crypto-secp256k1 module anyway, so no loss there.

BTW, I ended up not using rewire, since it wouldn't let me modify the `const keys` in index.js - instead I'm just modifying the object directly in the `before` and `after` for the test.  Since mocha tests are run sequentially this seems safe enough, but I could switch to rewire and change `keys` to be `let` instead of `const` if you'd prefer.

parkan:
Anything remains here?

dignifiedquire:
Sorry I forgot about this, I will review in the morning

dignifiedquire:
Thank you :octocat:

PR #74, opened 2017-02-11
refactor: loading of additional key support

dignifiedquire:


dignifiedquire:
@yusefnapora @diasdavid please review this

diasdavid:
This means that the extra set of keys needs to be 'injected' everywhere libp2p-crypto is required, which might be ok if the extra set of keys is not doing anything inside libp2p and only is used for the application using libp2p

yusefnapora:
I haven't had time to install this locally and play with it, but it seems like a good move to me.  I wasn't super keen on the hardcoded "require in a `try/catch` block" approach from the initial PR.  I especially like that you inject libp2p-crypto into the key modules so they don't need to depend on it directly.

I think it's fine to have users of the optional key types like secp256k1 jump through the `addKeyType` hoop; it feels less "magic" and opaque than having the key type suddenly available if the key module exists.

diasdavid:
@yusefnapora what might be the dealbreaker with this approach is that you won't be able to pick up any of the references to libp2p-crypto that exist inside libp2p and add the extra keys once and expect them to exist everywhere. This is caused because the way npm and semver worked, if two modules happen to depend in different versions, npm will create two different singletons, the solution for this is dependency injection all the way down, which is cumbersome.

In the other hand, if you don't care if the libp2p stack has access to this curves, it raises the question of why we are adding them here in the first place

yusefnapora:
@diasdavid that's not really an issue for us; we mostly just want our various key types to have a common interface, and having them be part of libp2p-crypto simplifies things.  Especially when it comes to marshaling/unmarshaling, since we pass around marshaled keys in a few places and using libp2p-crypto to unmarshal them.  Long-term we'd like to use something like multikey for sending self-identifying keys around, but it's not quite baked yet.

dignifiedquire:
out of date

Issue #84, opened 2017-03-01
Missing GenerateKeyPairWithReader


jackkleeman:
The js implementation lacks the ability to produce keys deterministically, as GenerateKeyPairWithReader in the Go implementation does. Ideally, js-libp2p-crypto would have a function to generate the same key given a seed as the Go implementation would, but at worst any kind of deterministic function would be useful. I am trying to port OpenBazaar to javascript, and currently the best I can do for deterministic key generation from a mnemonic is to have the Go implementation run through GopherJS, which is very slow.

dignifiedquire:
As we use the webcrypto api when available as far as I know we will not be able to provide seeds to do deterministic key generation. The same goes for the nodejs crypto, I'm not seeing any way to do deterministic generation there. 

What do you need deterministic keys for, testing?

JustinDrake:
Deterministic identities are used in OpenBazaar. The user remembers a single mnemonic to deterministically generate both a Bitcoin wallet and a peer id.

dignifiedquire:
Thanks for the info. I will try and figure out what possibilites there are doing this in JavaScript. It will probably be a stand alone tool though, to keep the size of this module as small as possible.

jackkleeman:
That would still be ideal!

dignifiedquire:
So using the latest version of node-forge I was able to do deterministic generation of keys. 

```js
'use strict'

const forge = require('node-forge')
const deepEqual = require('deep-equal')

forge.options.usePureJavaScript = true

const prng = () => {
  const p = forge.random.createInstance()
  p.seedFileSync = function(needed) {
    return forge.util.fillString('a', needed)
  }
  return p
}

const k1 = forge.rsa.generateKeyPair(512, {prng: prng()})
const k2 = forge.rsa.generateKeyPair(512, {prng: prng()})


console.log('same keys', deepEqual(k1.publicKey.n, k2.publicKey.n))
// => true
```

jackkleeman:
Fantastic! What we have found difficult is doing the same for elliptic curve identities, which OB uses. Ideally, we could have the same keys coming from the same seeds in JS and in Go.

jackkleeman:
It seems the elliptic curve library used by go-libp2p generates keys in the same way as tweetnacl, except that go-libp2p first applies scrypt to the seed. I am working on some js code which will produce the same key as go-libp2p given the same seed.

dignifiedquire:
[...]

PR #85, opened 2017-03-03
Implement generateKeyPairFromSeed for ed25519

jackkleeman:
This will produce the same keypair for the same seed (or first 32 bytes of reader) as given to GenerateKeyPairWithReader in go-libp2p. Solves #84 for ed25519.

### DIFF
--- a/test/ed25519.spec.js
+++ b/test/ed25519.spec.js
@@ -35,6 +35,27 @@ describe('ed25519', () => {
     })
   })
 
+  it('generates a valid key from seed', (done) => {
+    var seed = crypto.randomBytes(32)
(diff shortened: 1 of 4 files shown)


test/ed25519.spec.js
```diff
@@ -35,6 +35,27 @@ describe('ed25519', () => {
     })
   })
 
+  it('generates a valid key from seed', (done) => {
```
dignifiedquire: Could you please add a test that generating twice from the same seed actually results in the same key?

jackkleeman: Done!

dignifiedquire:
Thank you :octocat:

PR #98, opened 2017-04-07
Allow base64 to be left 0 padded, and do so in ecdh key unmarshalling

### LINKED ISSUE #97: ECDH Buffer to BN conversion loses leading 0s which are needed in webcrypto on some browsers
I have verified here:
https://runkit.com/jackkleeman/58e63d3546216b001451a645
That a buffer will lose a leading 0 byte when converted to a BN, and when this is converted to base64 for use in JWK format, this base64 will also have no leading 0. This is a problem because chrome rejects base64 jwk x and y values which don't have the full 32 bytes (silly I know):
https://chromium.googlesource.com/chromium/src/+/master/components/test/data/webcrypto/bad_ec_keys.json#138
Therefore, 1 in 256 (or possibly 1/128) keys will be invalid when [unmarshalled to JWK](https://github.com/libp2p/js-libp2p-crypto/blob/master/src/crypto/ecdh-browser.js#L114) , because their base64 string will not have the leading 0 byte. I am not sure what the simplest way to fix this is (maybe an optional pad-to-length input to the toBase64 util function?), although a test case is fairly trivial, generate a bunch of ecdh keys until one has a leading 0 byte.

cc @JustinDrake good job for catching this

jackkleeman:
JWK format, at least for elliptical key import, requires x and y values (base64 encoded) to be an expected byte length, so the buffer that comes out of the bn.js toArrayLike function must be left padded. Currently if the first byte of 32 is 0, a 31 byte buffer is returned from toArrayLike, and the resulting base64 reflects this. If 32 is passed to the toBase64() function as 'len', then the first byte is now set as 0.

This fixes #97. There may be similar issues elsewhere, in which case they can also pass a length value to toBase64(). In the meantime, this fixes unmarshalPublicKey() in ecdh-browser.js

### DIFF
--- a/src/crypto/ecdh-browser.js
+++ b/src/crypto/ecdh-browser.js
@@ -117,8 +117,8 @@ function unmarshalPublicKey (curve, key) {
   return {
     kty: 'EC',
     crv: curve,
-    x: toBase64(x),
-    y: toBase64(y),
+    x: toBase64(x, byteLen),
+    y: toBase64(y, byteLen),
     ext: true
   }
 }
--- a/src/crypto/util.js
+++ b/src/crypto/util.js
@@ -5,8 +5,9 @@ const Buffer = require('safe-buffer').Buffer
 
 // Convert a BN.js instance to a base64 encoded string without padding
 // Adapted from https://tools.ietf.org/html/draft-ietf-jose-json-web-signature-41#appendix-C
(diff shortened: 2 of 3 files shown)


dignifiedquire:
Thank you for catching and fixing this!

PR #103, opened 2017-08-16
Skip nextTick in nodeify

dignifiedquire:
We don't need it in our usage, as the promises we are transforming are always async.

Ref https://github.com/ipfs/js-ipfs/issues/128

### DIFF
--- a/src/nodeify.js
+++ b/src/nodeify.js
@@ -1,0 +1,11 @@
+'use strict'
+
+// Based on npmjs.com/nodeify but without additional `nextTick` calls
+// to keep the overhead low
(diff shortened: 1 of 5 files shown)


src/nodeify.js
```diff
@@ -0,0 +1,11 @@
...
+
+// Based on npmjs.com/nodeify but without additional `nextTick` calls
+// to keep the overhead low
+module.exports = function nodeify (promise, cb) {
+  return promise.then((res) => {
+    cb(null, res)
+  }, (err) => {
+    cb(err)
+  })
+}
```
diasdavid: 👍

PR #104, opened 2017-08-17
feat: use aes-js for faaaster crypto in the browser

dignifiedquire:
This doesn't have the stream overhead, and in general performs much better than browserify-aes

### DIFF
--- a/package.json
+++ b/package.json
@@ -6,7 +6,7 @@
   "browser": {
     "./src/hmac/index.js": "./src/hmac/index-browser.js",
     "./src/keys/ecdh.js": "./src/keys/ecdh-browser.js",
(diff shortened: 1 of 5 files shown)


diasdavid:
can we have some benchmarks? I bet it really is faster, I just want to know how much :)

dignifiedquire:
@diasdavid we don't have a good way to setup browser benchmarks atm, but you can look at https://github.com/ipfs/js-ipfs/issues/128#issuecomment-323103088 to see the after picture of applying this patch in real world usage

dignifiedquire:
Hold off on merging this please, might not be as good as I hoped

dignifiedquire:
Turns out this is actually not that much faster :(

PR #107, opened 2017-09-04
fix: switch protobufjs

dignifiedquire:
no more eval


.gitignore
```diff
@@ -43,3 +43,4 @@ test/test-data/go-ipfs-repo/LOG.old
 
 # while testing npm5
 package-lock.json
+yarn.lock
```
diasdavid: same here (gitignore the built protobuf)

PR #109, opened 2017-09-06
fix: use regular protocol-buffers until protobufjs is fixed

diasdavid:
interim fix for https://github.com/libp2p/js-libp2p-secio/issues/92#issuecomment-327261444

//cc @ya7ya


diasdavid:
Circle is drunk today 
image

Issue #112, opened 2017-11-27
Catch error when unmarshaling instead of crashing


mkg20001:
The `unmarshalPrivateKey` function throws an error instead of passing it to the callback which makes the app crash. (That might not be the only function with that bug)

Related libp2p/js-peer-id#72

mkg20001:
PR #113

diasdavid:
Continue on #113

PR #113, opened 2017-11-27
Catch error when unmarshaling instead of crashing

mkg20001:
Related #112

### DIFF
--- a/src/keys/rsa.js
+++ b/src/keys/rsa.js
@@ -9,43 +9,34 @@ const jwkToPem = require('pem-jwk').jwk2pem
 exports.utils = require('./rsa-utils')
 
 exports.generateKey = function (bits, callback) {
-  const done = (err, res) => setImmediate(() => callback(err, res))
-
-  let key
-  try {
-    key = keypair({ bits: bits })
-  } catch (err) {
-    return done(err)
-  }
-
-  let res
-  try {
-    res = {
-      privateKey: pemToJwk(key.private),
-      publicKey: pemToJwk(key.public)
+  setImmediate(() => {
+    let result
+    try {
+      const key = keypair({ bits: bits })
+      result = {
+        privateKey: pemToJwk(key.private),
+        publicKey: pemToJwk(key.public)
+      }
+    } catch (err) {
+      return callback(err)
     }
-  } catch (err) {
-    return done(err)
-  }
 
-  done(null, res)
+    callback(null, result)
+  })
 }
 
 // Takes a jwk key
 exports.unmarshalPrivateKey = function (key, callback) {
-  try {
-    key = {
+  setImmediate(() => {
+    callback(null, {
       privateKey: key,
       publicKey: {
         kty: key.kty,
         n: key.n,
         e: key.e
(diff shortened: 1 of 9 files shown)


mkg20001:
weird test results
just noticed that the rsa `key.verify` function doesn't return an error even when the key is invalid. is this intended?

src/keys/rsa.js
```diff
@@ -18,22 +18,34 @@ exports.generateKey = function (bits, callback) {
...
+      privateKey: key,
+      publicKey: {
+        kty: key.kty,
+        n: key.n,
+        e: key.e
+      }
     }
-  })
+  } catch (err) {
+    callback(new Error('Key is invalid!'))
```
diasdavid: needs a `return`

src/keys/rsa.js
```diff
@@ -18,22 +18,34 @@ exports.generateKey = function (bits, callback) {
...
-      n: key.n,
-      e: key.e
+  try {
+    key = {
+      privateKey: key,
+      publicKey: {
+        kty: key.kty,
+        n: key.n,
+        e: key.e
+      }
```
diasdavid: Where would an exception be thrown in this operation?

test/testCb.js
```diff
@@ -0,0 +1,42 @@
...
+      args.push(garbage)
+    }
+    it(fncName + '(' + args.map(garbage => util.inspect(garbage)).join(', ') + ')', cb => {
+      args.push((err, res) => {
+        expect(err).to.exist()
+        expect(res).to.not.exist()
+        cb()
+      })
+
+      fnc(...args)
```
diasdavid: Please do not use this operator for now (we are still avoiding some ES6 features)

src/keys/rsa.js
mkg20001: Never underestimate the power of undefined: `TypeError: Cannot read property 'e' of undefined`

diasdavid: True, but in this case, it would never through that error, only if you did something like `key.e.something`. As long as if `key` exists and it is an object, `key.e` will never throw

pgte:
Going to push some changes here to help this move along.

src/keys/rsa.js
pgte: reverted this, but now inside a `setImmediate`.

pgte:
@diasdavid should be go to go. Please tell if there's anything else you'd like to see addressed here.

src/keys/rsa.js
mkg20001: And what about `unmarshalPrivateKey(undefined, (err, res) => {})` ??

diasdavid:
Thanks @pgte :) @VictorBjelkholm could you check what's wrong with Circle?

src/keys/rsa.js
pgte: @mkg20001 good point, fixing.

diasdavid:
@VictorBjelkholm seems that ci-sync never got here. Mind checking?

Going to merge for now.

PR #117, opened 2017-12-25
Improve speed

richardschneider:
`jsrsasign` is 2 orders of magnitude slower that `node-forge`; see the [benchmarks](https://github.com/richardschneider/crypto-benchmarks).

[node-forge](https://unpkg.com/node-forge/dist/) is a little bigger than [jsrsasign](https://unpkg.com/jsrsasign/lib/) (273 kB vs 259 kB).

`libp2p-crypto` relies upon `keypair` for RSA.  It is 128 kB.  `keypair` has copied `node-forge` code from 5 years ago and has added JWK support.  If we make a JWK PR to forge, then we can remove the the `keypair` dependency.

### DIFF
--- a/src/keys/rsa-class.js
+++ b/src/keys/rsa-class.js
@@ -121,7 +121,6 @@ class RsaPrivateKey {
    * @returns {undefined}
    */
   export (format, password, callback) {
-    const self = this
     if (typeof password === 'function') {
       callback = password
       password = format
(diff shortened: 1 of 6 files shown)


diasdavid:
@richardschneider have you tried npm linking this package to its dependents and run the tests? The tricky part with UMD comes appears when you try to require a module that has a dependency using it.

richardschneider:
`forge` is now CommonJS so UMD issues disappear.  All tests pass on all platforms and CIs. 

Not sure what you want to `npm link`?  I have npm link crypto to keychain and it works.

diasdavid:
Interesting, that seems to be the case now https://www.npmjs.com/package/node-forge#installation ✨

I'll test this ASAP

richardschneider:
@diasdavid Thanks.  IMHO node-forge is a better package than `jsrsasign`

richardschneider:
@diasdavid LGTM

src/keys/rsa-class.js
```diff
@@ -121,21 +121,30 @@ class RsaPrivateKey {
    * @returns {undefined}
    */
   export (format, password, callback) {
+    const self = this
```
diasdavid: Never gets used.

PR #128, opened 2018-07-29
feat: use ursa-optional for lightning fast key generation

mkg20001:
The difference between ursa and ursa-optional is that ursa-optional does not cause any problems if it fails to compile

### DIFF
--- a/package.json
+++ b/package.json
@@ -42,7 +42,7 @@
     "protons": "^1.0.1",
     "rsa-pem-to-jwk": "^1.1.3",
     "tweetnacl": "^1.0.0",
-    "ursa-optional": "^0.9.8",
+    "ursa-optional": "~0.9.8",
     "webcrypto-shim": "github:dignifiedquire/webcrypto-shim#master"
   },
   "devDependencies": {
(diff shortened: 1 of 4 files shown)


diasdavid:
@dignifiedquire can you verify, merge and release?

diasdavid:
@mkg20001 can you bring coverage up?

mkg20001:
@diasdavid Added tests that run .generateKeyPair() with both crypto libs

mkg20001:
As I neither have a windows machine nor VM currently, somebody else would have to debug whats causing this issue. Most likely the module just failed to compíle.

mkg20001:
@dignifiedquire Could you merge this?

package.json
```diff
@@ -42,6 +42,7 @@
     "protons": "^1.0.1",
     "rsa-pem-to-jwk": "^1.1.3",
     "tweetnacl": "^1.0.0",
+    "ursa-optional": "^0.9.8",
```
diasdavid: This should be failing the linting. Use ~ instead of ^ for < 1.0.0 modules.

dignifiedquire:
Sorry, this got lost in my inbox :(

PR #131, opened 2018-10-27
feat: use async await

alanshaw:
This PR changes this module to remove callbacks and use async/await. The API is unchanged aside from the obvious removal of the `callback` parameter.

### `async` keyword

I've labeled API methods that are asynchronous with the `async` keyword even if they don't actually use `await`. I'm interested to hear views on this but I think it communicates the intention of the function better even if it is not required. A contrived example:

```js
async function publicApiFn (type) {
  type = type.toLowerCase()
  return _privateFn(type) // an async function / returns a promise
}
```

`publicApiFn` doesn't actually use `await`, so `async` is not strictly required. However the `async` keyword helps communicate the fact that it returns a promise so you can use it without having to track down `_privateFn` to see what that returns.

The other reason for doing this is if a consumer is using `then/catch` and `type.toLowerCase()` throws, then it'll be caught in the catch block. Without the `async` keyword it'll be an uncaught exception.

### Mostly synchronous API

There's a lot of functions labelled as async that don't actually do anything async! This is largely because the Node.js crypto API is synchronous. However the webcrypto API is async so I assume this has been done to present a consistent interface to developers in both Node.js and the browser. Maybe someone can verify that for me?

---

depends on

* [x] https://github.com/multiformats/js-multihashing-async/pull/37
* [x] https://github.com/libp2p/js-libp2p-crypto-secp256k1/pull/9

refs https://github.com/ipfs/js-ipfs/issues/1670

---

# Benchmarks

## Before

```
$ node benchmarks/ephemeral-keys.js 
ephemeral key with secrect P-256 x 5,825 ops/sec ±2.25% (74 runs sampled)
ephemeral key with secrect P-384 x 5,938 ops/sec ±3.78% (77 runs sampled)
ephemeral key with secrect P-521 x 6,044 ops/sec ±2.72% (77 runs sampled)

$ node benchmarks/key-stretcher.js 
keyStretcher AES-128 SHA1 x 9,810 ops/sec ±3.23% (72 runs sampled)
keyStretcher AES-128 SHA256 x 13,401 ops/sec ±4.01% (74 runs sampled)
keyStretcher AES-128 SHA512 x 22,072 ops/sec ±4.63% (70 runs sampled)
keyStretcher AES-256 SHA1 x 8,631 ops/sec ±2.72% (76 runs sampled)
keyStretcher AES-256 SHA256 x 10,794 ops/sec ±5.49% (71 runs sampled)
keyStretcher AES-256 SHA512 x 16,027 ops/sec ±6.36% (74 runs sampled)
keyStretcher Blowfish SHA1 x 81,146 ops/sec ±9.82% (71 runs sampled)
keyStretcher Blowfish SHA256 x 75,919 ops/sec ±13.66% (72 runs sampled)
keyStretcher Blowfish SHA512 x 70,505 ops/sec ±20.75% (67 runs sampled)

$ node benchmarks/rsa.js 
generateKeyPair 1024bits x 73.18 ops/sec ±6.67% (53 runs sampled)
generateKeyPair 2048bits x 16.51 ops/sec ±15.23% (30 runs sampled)
generateKeyPair 4096bits x 1.89 ops/sec ±47.81% (14 runs sampled)
sign and verify x 1,887 ops/sec ±1.77% (73 runs sampled)
```

## After

```
$ node benchmarks/ephemeral-keys.js 
ephemeral key with secrect P-256 x 6,275 ops/sec ±2.17% (76 runs sampled)
ephemeral key with secrect P-384 x 6,291 ops/sec ±3.32% (78 runs sampled)
ephemeral key with secrect P-521 x 6,272 ops/sec ±3.35% (76 runs sampled)

$ node benchmarks/key-stretcher.js 
keyStretcher AES-128 SHA1 x 16,948 ops/sec ±1.23% (80 runs sampled)
keyStretcher AES-128 SHA256 x 22,188 ops/sec ±1.48% (76 runs sampled)
keyStretcher AES-128 SHA512 x 34,204 ops/sec ±2.26% (77 runs sampled)
keyStretcher AES-256 SHA1 x 14,910 ops/sec ±0.85% (82 runs sampled)
keyStretcher AES-256 SHA256 x 17,841 ops/sec ±1.75% (77 runs sampled)
keyStretcher AES-256 SHA512 x 26,028 ops/sec ±1.57% (73 runs sampled)
keyStretcher Blowfish SHA1 x 113,697 ops/sec ±8.02% (75 runs sampled)
keyStretcher Blowfish SHA256 x 93,089 ops/sec ±28.10% (55 runs sampled)
keyStretcher Blowfish SHA512 x 34,725 ops/sec ±50.76% (29 runs sampled)

$ node benchmarks/rsa.js 
generateKeyPair 1024bits x 71.01 ops/sec ±7.37% (48 runs sampled)
generateKeyPair 2048bits x 13.97 ops/sec ±10.97% (34 runs sampled)
generateKeyPair 4096bits x 2.59 ops/sec ±29.13% (17 runs sampled)
sign and verify x 1,989 ops/sec ±5.39% (79 runs sampled)
```


alanshaw:
@mkg20001 any idea why `ursa-optional` is failing to install on windows now? We get no error message :(

mkg20001:
@alanshaw The error messages usually get stored in `node_modules/ursa-optional/stderr.log` and `node_modules/ursa-optional/stdout.log`.
I'd love to debug that but sadly I'm stuck at a place with an internet connection so slow it takes _up to 10 seconds_ to SSH into anything and the only windows machine I have is a VM on a server.
At least ursa-optional is optional so it's not that much of a problem.

alanshaw:
Ah, thanks...we can't look at them in CI though. Is there a reason why they're not just sent to stdout and stderr? Makes for nice npm install output but difficult to debug from here.

mkg20001:
@coolaj86 suggested to add that as people might see the error message and think the thing is broken while in fact it's not. I'll release a new version with an env flag `URSA_VERBOSE` which would print the output directly if set, that would have to be enabled on CI, too

diasdavid:
> This is largely because the Node.js crypto API is synchronous. However the webcrypto API is async so I assume this has been done to present a consistent interface to developers in both Node.js and the browser. Maybe someone can verify that for me?

This is exactly correct. We need to keep things `async` otherwise we would have to interact with both the Node.js Crypto and WebCrypto differently changing a bunch of modules upstream.

dignifiedquire:
@alanshaw what is the state of this?

achingbrain:
I've updated the commit message

src/keys/rsa-class.js
```diff
@@ -3,22 +3,18 @@
 const multihashing = require('multihashing-async')
 const protobuf = require('protons')
 const bs58 = require('bs58')
-const nextTick = require('async/nextTick')
 
 const crypto = require('./rsa')
 const pbm = protobuf(require('./keys.proto'))
-require('node-forge/lib/sha512')
```
vasco-santos: I believe @hugomrdias added this for reducing the bundle size: https://github.com/libp2p/js-libp2p-crypto/commit/8d8294dc3f9b15d827536f5f79fb512eaf4384a9#diff-17ddd891dd4a459b2c549aebf399e5edR10

Should we really remove this? If not, we should document why we have this

achingbrain: Good catch, removing this increased the bundle size by 400kb(!)(!!!!!!!).  I have put it back in.

README.md
```diff
@@ -56,130 +59,121 @@ Expoes an interface to AES encryption (formerly Rijndael), as defined in U.S. Fe
...
 Installing the [libp2p-crypto-secp256k1](https://github.com/libp2p/js-libp2p-crypto-secp256k1) module adds support for the `'secp256k1'` type, which supports ECDSA signatures using the secp256k1 elliptic curve popularized by Bitcoin.  This module is not installed by default, and should be explicitly depended on if your project requires secp256k1 support.
 
-### `crypto.keys.generateKeyPair(type, bits, callback)`
+### `crypto.keys.generateKeyPair(type, bits)`
 
 - `type: String`, see [Supported Key Types](#supported-key-types) above.
 - `bits: Number` Minimum of 1024
-- `callback: Function`
+
+Returns `Promise<privateKey<Buffer>, publicKey<Buffer>>`
```
dirkmc: Does the Promise resolve to an array or an object?

README.md
```diff
@@ -214,45 +209,55 @@ Calls back with an object of the form:
 }
 ```
 
-### `crypto.keys.marshalPublicKey(key[, type], callback)`
+### `crypto.keys.marshalPublicKey(key, [type])`
 
 - `key: keys.rsa.RsaPublicKey | keys.ed25519.Ed25519PublicKey | require('libp2p-crypto-secp256k1').Secp256k1PublicKey`
 - `type: String`, see [Supported Key Types](#supported-key-types) above.
```
dirkmc: What does type default to if not specified?

src/hmac/index.js
```diff
@@ -2,21 +2,16 @@
...
 const crypto = require('crypto')
 const lengths = require('./lengths')
-const nextTick = require('async/nextTick')
 
-exports.create = function (hash, secret, callback) {
+exports.create = async function (hash, secret) { // eslint-disable-line require-await
   const res = {
-    digest (data, cb) {
+    async digest (data) { // eslint-disable-line require-await
       const hmac = crypto.createHmac(hash.toLowerCase(), secret)
```
dirkmc: Why do we lowercase the string here?

src/keys/ecdh.js
```diff
@@ -1,41 +1,28 @@
...
-const nextTick = require('async/nextTick')
 
 const curves = {
   'P-256': 'prime256v1',
   'P-384': 'secp384r1',
   'P-521': 'secp521r1'
 }
 
-exports.generateEphmeralKeyPair = function (curve, callback) {
+exports.generateEphmeralKeyPair = async function (curve) { // eslint-disable-line require-await
```
dirkmc: Does this need to be an async function?

src/keys/ed25519-class.js
```diff
@@ -86,81 +82,34 @@ class Ed25519PrivateKey {
...
-    }
-
-    cb(null, privkey)
-  })
+async function generateKeyPair () {
+  const { secretKey, publicKey } = await crypto.generateKey()
+  return new Ed25519PrivateKey(secretKey, publicKey)
 }
 
-function generateKeyPairFromSeed (seed, _bits, cb) {
```
dirkmc: Wow that is a lot less LOC! :)

src/keys/key-stretcher.js
```diff
@@ -37,72 +36,38 @@ module.exports = (cipherType, hash, secret, callback) => {
...
-            cb()
-          })
-        })
-      }
-
-      function finish (err) {
-        if (err) {
-          return callback(err)
-        }
+    if (j + todo > resultLength) {
```
dirkmc: I'm not sure if this if block is necessary?

src/keys/rsa-class.js
```diff
@@ -102,104 +93,67 @@ class RsaPrivateKey {
...
    * @param {string} [format] - Defaults to 'pkcs-8'.
    * @param {string} password - The password to read the encrypted PEM
-   * @param {function(Error, KeyInfo)} callback
-   * @returns {undefined}
+   * @returns {KeyInfo}
    */
-  export (format, password, callback) {
-    if (typeof password === 'function') {
-      callback = password
+  async export (format, password) { // eslint-disable-line require-await
```
dirkmc: Given that format is optional, would it make sense to reverse the argument order?

src/hmac/index.js
dirkmc: Nevermind I see :)

README.md
achingbrain: Lol comments as a type system.

I think it's supposed to be an object with `privateKey` and `publicKey` properties, both of which are buffers.

README.md
achingbrain: `rsa`, will update the docs

README.md
achingbrain: Will update the docs

src/hmac/index.js
achingbrain: From what I can see, our docs have things like `SHA512` as values for this field but the `crypto` module expects it to be `sha512`.

src/keys/ecdh.js
achingbrain: The browser implementation is async but the node implementation is sync.  I think the intention here is to present the same interface.  See @alanshaw's note titled 'async keyword' in the PR description.

src/keys/rsa-class.js
achingbrain: Yes, but it'd be good to limit the scope of this PR to converting the API to async/await rather than general API improvements.

src/keys/key-stretcher.js
achingbrain: Certainly removing it doesn't make any tests fail, though it's in the callback version too.

README.md
achingbrain: Done!

README.md
achingbrain: Done!

src/keys/rsa-class.js
achingbrain: I've created an issue for this: #153

jacobheun:
@dignifiedquire I am going to merge this in so that we can incorporate #153 as well before the release. If there are any issues you see with this we can do a follow up PR before the release.

Issue #46, opened 2016-12-02
WebCrypto.getRandomValues vs crypto.randomBytes


diasdavid:
#### crypto.randomBytes

https://nodejs.org/api/crypto.html#crypto_crypto_randombytes_size_callback

returns a Buffer with random bytes given the size

#### WebCrypto.getRandomValues(arr)

https://developer.mozilla.org/en/docs/Web/API/RandomSource/getRandomValues

overwrites an array with randomBytes

dignifiedquire:
are you trying to collect all strange behavior of webcrypto here? ;)

diasdavid:
@dignifiedquire exactly, hopefully it will be useful to enumerate issues in the future

dignifiedquire:
As a side note, the webcrypto api is much more sensible when trying to optimize for speed and allocations. You can manage your own array and fill it with random values, whereas the node api is actually much worse, as it forces allocation of a new buffer when generating random values.

diasdavid:
Still a `WAT`

daviddahl:
See: https://bugzilla.mozilla.org/show_bug.cgi?id=440046#c39 

and

https://bugzilla.mozilla.org/show_bug.cgi?id=673432#c5

PR #134, opened 2018-12-05
fix: clean up, bundle size reduction

hugomrdias:
This PR cleanup the repo with some linting, package.json updates, etc, plus does the first pass and reducing bundle size (OKR work).
Also proper isomorphic randomBytes is added.

bundle size:  175KB > 138.85KB **-20.65%**

Closes https://github.com/libp2p/js-libp2p-crypto/issues/46

Works better with this https://github.com/ipfs/aegir/pull/305


src/keys/rsa-browser.js
```diff
@@ -49,23 +49,21 @@ exports.unmarshalPrivateKey = function (key, callback) {
   })), callback)
 }
 
-exports.getRandomValues = function (arr) {
-  return Buffer.from(webcrypto.getRandomValues(arr))
-}
+exports.getRandomValues = randomBytes
```
jacobheun: getRandomValues calls need to be fixed. You've changed the fingerprint on the param from array to number. There's also no guarantee consumers of libp2p-crypto aren't using these methods. If it's going to change, imo, we should do some type checking and then remove support for array in a future release.

hugomrdias: These aren't documented right? we should be ok to change. 
The only one documented is this one https://github.com/libp2p/js-libp2p-crypto#cryptorandombytesnumber. 

also rsa.js and rsa-browser.js basically feed top level randomBytes

jacobheun: I think we should at least note this as a breaking change in the commit for changelog, in case someone is using it, it will be easy to see the change. The tests are currently failing because of the switch from array to number. Internal usage needs to be fixed.

hugomrdias: ups let me check

jacobheun:
Linting needs to be fixed, otherwise this looks good.

src/random-bytes.js
```diff
@@ -1,13 +1,9 @@
...
-module.exports = randomBytes
+'use strict';
+const randomBytes = require('iso-random-stream/src/random');
+
+module.exports = function (number) {
+    if (!number || typeof number !== 'number') {
+        throw new Error('first argument must be a Number bigger than 0');
+    }
+    return randomBytes(number);
+};
```
vasco-santos: There are some issues in this file with linting

src/keys/rsa-browser.js
dignifiedquire: please leave the array behavior intact. this is making the api less efficient than it has to be, as this means the caller is forced to allocate a new array instead of reusing an existing one

package.json
```diff
@@ -34,27 +39,28 @@
...
-    "libp2p-crypto-secp256k1": "~0.2.2",
+    "libp2p-crypto-secp256k1": "libp2p/js-libp2p-crypto-secp256k1#fix/bundle-size",
     "multihashing-async": "~0.5.1",
     "node-forge": "~0.7.6",
     "pem-jwk": "^1.5.1",
     "protons": "^1.0.1",
     "rsa-pem-to-jwk": "^1.1.3",
     "tweetnacl": "^1.0.0",
-    "ursa-optional": "~0.9.9",
-    "webcrypto-shim": "github:dignifiedquire/webcrypto-shim#master"
```
dignifiedquire: are you sure you can just drop these shims? I don’t think the state of webcrypto has improved much since this was written

src/keys/rsa-browser.js
hugomrdias: these methods are not used nor documented, can we look into this in another issue/PR ? i can easily add support for passing arrays here https://github.com/hugomrdias/iso-random-stream/blob/master/src/random.browser.js if we need to

package.json
hugomrdias: for the current support/target browserlist yes we can

hugomrdias:
@alanshaw ready to merge/release

PR #150, opened 2019-07-09
feat: bring your own web crypto

alanshaw:
Changes `webcrypto.js` to check for native web crypto availability and falls back to using `window.__crypto` if not available.

If the user wants to bring their own Web Crypto API compatible implementation then they simply need to assign it to `window.__crypto` before they start using IPFS/libp2p.

Checks are done in the functions that require web crypto to give the user the flexibility to assign to `window.__crypto` before OR after they import `libp2p-crypto`. It also means that users have the ability to use other exported functions that do not require web crypto without having to worry about sorting their own implementation.

We use `window.__crypto` because `window.crypto` is a readonly property in secure context and always readonly in workers.

If `window.crypto` and `window.__cypto` are unavailable then an appropriate error message is reported to the user with a `ERR_MISSING_WEB_CRYPTO` code.

I've also added documentation to the README.

This is a backwards compatible change.

closes https://github.com/libp2p/js-libp2p-crypto/pull/149
resolves https://github.com/libp2p/js-libp2p-crypto/issues/105
resolves https://github.com/ipfs/js-ipfs/issues/2017
resolves https://github.com/ipfs/js-ipfs/issues/2153

cc @lidel

### DIFF
--- a/README.md
+++ b/README.md
@@ -48,11 +48,40 @@ This repo contains the JavaScript implementation of the crypto primitives needed
 npm install --save libp2p-crypto
 ```
 
+## Usage
+
+```js
+const crypto = require('libp2p-crypto')
+
+// Now available to you:
+//
+// crypto.aes
+// crypto.hmac
+// crypto.keys
+// etc.
+//
+// See full API details below...
+```
+
+### Web Crypto API
+
+The `libp2p-crypto` library depends on the [Web Crypto API](https://developer.mozilla.org/en-US/docs/Web/API/Web_Crypto_API) in the browser. Web Crypto is available in all modern browsers, however browsers restrict its usage to [Secure Contexts](https://developer.mozilla.org/en-US/docs/Web/Security/Secure_Contexts).
+
+**This means you will not be able to use `libp2p-crypto` in the browser when the page is served over HTTP.**
+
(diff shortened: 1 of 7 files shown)


achingbrain:
What's the thinking behind requiring the user to modify global state instead of passing a crypto implementation in as an option?

daviddias:
I share @achingbrain's question. This will turn into a "very obscure and hard to understand how it works" hack. 

I suggest doing the long way, passing the crypto module as a libp2p constructor option.

jacobheun:
A lot of modules leverage libp2p-crypto directly instead of pulling from libp2p itself. In order to pass crypto to libp2p and have the modules leverage it, its usage will need to change everywhere. DI could make this much easier to resolve.

alanshaw:
Thinking behind:

Not to break changes.

Keep scope low.

Fix it soon.

You’d have to pass the crypto impl to each function since this module isn’t a class you instantiate it’s a grab bag of crypto functions. 

Or refactor the module internally to become a class and somehow thread that through lots of other libp2p modules with DI or pass the config around everywhere.

Perhaps we could have a static setCryptoImpl on the module...but if we ever have two different versions in a bundle it won’t be shared.

I’m open to suggestions, or what about merging this and fixing it properly in the future?

This is what PRs are for 😅

PR #151, opened 2019-07-10
Upgrade libp2p-crypto-secp256k1 + fix Secp256k1 tests

aknuds1:
Upgrade libp2p-crypto-secp256k1 dependency and ensure that Secp256k1 supported is tested properly.

### DIFF
--- a/package.json
+++ b/package.json
@@ -57,7 +57,8 @@
     "bundlesize": "~0.18.0",
     "chai": "^4.2.0",
     "chai-string": "^1.5.0",
-    "dirty-chai": "^2.0.1"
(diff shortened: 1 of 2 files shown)


aknuds1:
@jacobheun @vasco-santos Here's a fix to ensure that Secp256k1 is supported through this library.

aknuds1:
> Fantastic! Since this is just adding tests I'd like to merge this in after #131. I can take on rebasing and updated this to async/await, once that's done.

@jacobheun Looking forward to seeing the new async API :D

jacobheun:
Rebased this with master and switched the tests over to async/await!

Issue #153, opened 2019-07-10
Make optional arguments last


achingbrain:
Methods like the `export` from `src/keys/rsa-class.js` have optional arguments first in their signature.  They should be last to follow conventions.

cc @dirkmc

dirkmc:
I'd suggest we do this immediately after the async changes, and before we do a release, as it's a breaking change

jacobheun:
I agree, since this needs to get pulled into several of the modules, we should make the change before going through that effort. 

The only caveat is that it will make those migrations to async/await a little more confusing in PRs moving the params around.

jacobheun:
async/await is in master, @achingbrain or @dirkmc, do either of you have time to do this? I'm working on a couple of other things but can take this after (likely tomorrow) if neither of you do.

dirkmc:
Sure I can do it

PR #154, opened 2019-07-10
fix: put optional args last for key export

dirkmc:
Fixes https://github.com/libp2p/js-libp2p-crypto/issues/153

Note: This probably isn't going to require any upstream code changes, because there is only one allowed export format and it's the default

### DIFF
--- a/test/keys/rsa.spec.js
+++ b/test/keys/rsa.spec.js
@@ -88,7 +88,7 @@ describe('RSA', function () {
 
   describe('export and import', () => {
     it('password protected PKCS #8', async () => {
-      const pem = await key.export('pkcs-8', 'my secret')
+      const pem = await key.export('my secret', 'pkcs-8')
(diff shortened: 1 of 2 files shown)


test/keys/rsa.spec.js
```diff
@@ -88,7 +88,7 @@ describe('RSA', function () {
 
   describe('export and import', () => {
     it('password protected PKCS #8', async () => {
-      const pem = await key.export('pkcs-8', 'my secret')
+      const pem = await key.export('my secret', 'pkcs-8')
```
vasco-santos: Is it worth adding a test without an explicit format?

dirkmc: There is one :) It's on line 98

vasco-santos: oh ok, didn't realize it!

Issue #152, opened 2019-07-10
Add `err-code` style codes to errors


achingbrain:
This module should not throw bare errors, instead they should be decorated with error codes using the [`err-code`](https://www.npmjs.com/package/err-code) module.

This improves our ability to do i18n and also lets us move away from asserting on error messages in tests, among other things.

PR #155, opened 2019-07-10
chore: add error codes

dirkmc:
Fixes https://github.com/libp2p/js-libp2p-crypto/issues/152

### DIFF
--- a/src/aes/cipher-mode.js
+++ b/src/aes/cipher-mode.js
@@ -1,0 +1,17 @@
+'use strict'
+
+const errcode = require('err-code')
+
+const CIPHER_MODES = {
+  16: 'aes-128-ctr',
+  32: 'aes-256-ctr'
+}
+
+module.exports = function (key) {
+  const mode = CIPHER_MODES[key.length]
+  if (!mode) {
+    const modes = Object.entries(CIPHER_MODES).map(([k, v]) => `${k} (${v})`).join(' / ')
+    throw errcode(`Invalid key length ${key.length} bytes. Must be ${modes}`, 'ERR_INVALID_KEY_LENGTH')
+  }
+  return mode
+}
--- a/src/aes/cipher-mode.js
+++ b/src/aes/cipher-mode.js
@@ -11,7 +11,7 @@ module.exports = function (key) {
   const mode = CIPHER_MODES[key.length]
   if (!mode) {
     const modes = Object.entries(CIPHER_MODES).map(([k, v]) => `${k} (${v})`).join(' / ')
-    throw errcode(`Invalid key length ${key.length} bytes. Must be ${modes}`, 'ERR_INVALID_KEY_LENGTH')
+    throw errcode(new Error(`Invalid key length ${key.length} bytes. Must be ${modes}`), 'ERR_INVALID_KEY_LENGTH')
   }
   return mode
 }
--- a/src/aes/index-browser.js
+++ b/src/aes/index-browser.js
@@ -1,11 +1,11 @@
 'use strict'
 
 const asm = require('asmcrypto.js')
-const cipherMode = require('./cipher-mode')
+const validateCipherMode = require('./cipher-mode')
 
 exports.create = async function (key, iv) { // eslint-disable-line require-await
-  // Check cipher mode
-  cipherMode(key)
+  // Throws an error if mode is invalid
+  validateCipherMode(key)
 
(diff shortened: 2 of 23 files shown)


src/aes/cipher-mode.js
```diff
@@ -0,0 +1,17 @@
...
+const CIPHER_MODES = {
+  16: 'aes-128-ctr',
+  32: 'aes-256-ctr'
+}
+
+module.exports = function (key) {
+  const mode = CIPHER_MODES[key.length]
+  if (!mode) {
+    const modes = Object.entries(CIPHER_MODES).map(([k, v]) => `${k} (${v})`).join(' / ')
+    throw errcode(`Invalid key length ${key.length} bytes. Must be ${modes}`, 'ERR_INVALID_KEY_LENGTH')
```
vasco-santos: Do the `new Error(__)` wrapping the error message here. Before we were not doing that, as the `err-code` module does that for us if we don't do it, but then the error stack will start from there

src/aes/index-browser.js
```diff
@@ -1,11 +1,11 @@
...
 
 const asm = require('asmcrypto.js')
+const cipherMode = require('./cipher-mode')
 
 exports.create = async function (key, iv) { // eslint-disable-line require-await
-  if (key.length !== 16 && key.length !== 32) {
-    throw new Error('Invalid key length')
-  }
+  // Check cipher mode
+  cipherMode(key)
```
vasco-santos: maybe we could name the function better, such as `validateCipherMode` in the `require`, so that the intended behavior is more understandable

src/keys/ecdh-browser.js
```diff
@@ -96,7 +97,7 @@ function unmarshalPublicKey (curve, key) {
   const byteLen = curveLengths[curve]
 
   if (!key.slice(0, 1).equals(Buffer.from([4]))) {
-    throw new Error('Invalid key format')
+    throw errcode('Cannot unmarshal public key - invalid key format', 'ERR_INVALID_KEY_FORMAT')
```
vasco-santos: use `new Error()`. This also happens in other places below, I will not point them out in this review

test/keys/ephemeral-keys.spec.js
```diff
@@ -60,4 +60,13 @@ describe('generateEphemeralKeyPair', () => {
       expect(secrets[0]).to.have.length(32)
     })
   })
+
+  it(`handles bad curve name`, async () => {
+    try {
+      await crypto.keys.generateEphemeralKeyPair('bad name')
+    } catch (err) {
+      return
```
achingbrain: Should probably assert on the error code here to avoid [Pokemon Exception Handling](https://blog.codinghorror.com/new-programming-jargon/)..

achingbrain: Sorry, should have added this comment yesterday.  Everything else LGTM

src/keys/rsa.js
```diff
@@ -40,7 +41,7 @@ exports.generateKey = async function (bits) { // eslint-disable-line require-awa
 // Takes a jwk key
 exports.unmarshalPrivateKey = async function (key) { // eslint-disable-line require-await
   if (!key) {
-    throw new Error('Key is invalid')
+    throw errcode(new Error('Mising key parameter'), 'ERR_MISSING_KEY')
```
jacobheun:
```suggestion
    throw errcode(new Error('Missing key parameter'), 'ERR_MISSING_KEY')
```

test/keys/ephemeral-keys.spec.js
dirkmc: Lucky you did - uncovered a bug, thanks!

achingbrain: Phew!

src/random-bytes.js
```diff
@@ -1,9 +1,10 @@
 'use strict'
 const randomBytes = require('iso-random-stream/src/random')
+const errcode = require('err-code')
 
-module.exports = function (number) {
-  if (!number || typeof number !== 'number') {
-    throw new Error('first argument must be a Number bigger than 0')
+module.exports = function (length) {
+  if (!length || typeof length !== 'number' || length <= 0) {
```
jacobheun:
```suggestion
  if (isNaN(length) || length <= 0) {
```

dirkmc: Thanks 👍

PR #157, opened 2019-07-17
fix: better error for missing web crypto

alanshaw:
This PR simply detects missing web crypto and throws an error with an appropriate message.

This is a stepping stone that will help users understand the problem until we have time to do a refactor of this module and of all the modules that use it to enable optionally passing your own crypto implementation.

refs https://github.com/libp2p/js-libp2p-crypto/pull/149
refs https://github.com/libp2p/js-libp2p-crypto/pull/150
refs https://github.com/libp2p/js-libp2p-crypto/issues/105
refs https://github.com/ipfs/js-ipfs/issues/2153
refs https://github.com/ipfs/js-ipfs/issues/2017

### DIFF
--- a/README.md
+++ b/README.md
@@ -51,11 +51,32 @@ This repo contains the JavaScript implementation of the crypto primitives needed
 npm install --save libp2p-crypto
 ```
 
+## Usage
+
+```js
+const crypto = require('libp2p-crypto')
+
+// Now available to you:
+//
+// crypto.aes
+// crypto.hmac
+// crypto.keys
+// etc.
+//
+// See full API details below...
+```
+
+### Web Crypto API
+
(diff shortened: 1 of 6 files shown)


README.md
```diff
@@ -51,11 +51,32 @@ This repo contains the JavaScript implementation of the crypto primitives needed
...
+// etc.
+//
+// See full API details below...
+```
+
+### Web Crypto API
+
+The `libp2p-crypto` library depends on the [Web Crypto API](https://developer.mozilla.org/en-US/docs/Web/API/Web_Crypto_API) in the browser. Web Crypto is available in all modern browsers, however browsers restrict its usage to [Secure Contexts](https://developer.mozilla.org/en-US/docs/Web/Security/Secure_Contexts).
+
+**This means you will not be able to use some `libp2p-crypto` functions in the browser when the page is served over HTTP.**
```
dirkmc: I suggest adding a couple of words to make it clear that you *can* serve over HTTPS

jacobheun:
I rebased this with the latest master so we can include this and the error code update in the next release.

alanshaw:
@dignifiedquire 🙏 can we get a merge and release?

alanshaw:
❤️ thanks @dignifiedquire

PR #158, opened 2019-09-23
V0.16.x upgrade node forge

achingbrain:
Upgrades `node-forge` dependency to the same one used by `libp2p-keychain` in order to downgrade `libp2p-keychain`'s `peer-id` dependency to the same version as the `libp2p` in order to upgrade `ipfs` to use the latest `libp2p` and still be able to dial other nodes on the network.

Phew.

This is necessary because otherwise `node-forge` complains that `sha512` is an unknown hash algorithm.

N.b. this PR is against the branch `v0.16.x` and not master, which is `v0.17.0` so needs to be released as `0.16.2`.

Also upgrades `aegir` to `v20` because for whatever reason the tests weren't running on browsers with `v18` any more.  It upgrades `standard` so there's some refactoring noise in this PR.

### DIFF
--- a/package.json
+++ b/package.json
@@ -45,7 +45,7 @@
     "keypair": "^1.0.1",
     "libp2p-crypto-secp256k1": "~0.3.0",
     "multihashing-async": "~0.5.1",
-    "node-forge": "~0.7.6",
+    "node-forge": "^0.8.5",
     "pem-jwk": "^2.0.0",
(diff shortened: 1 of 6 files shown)


vasco-santos:
@dignifiedquire can you release this as a `0.16.x` release? 🙏

dignifiedquire:
done

PR #161, opened 2020-01-11
Adds typescript types + linting/tests

carsonfarmer:
This preliminary PR adds Typescript types for libp2p-crypto. It includes linting-based 'tests', which were derived from existing libp2p-crypto tests. Is also includes one new dependency, `dtslint`, which is the test/lint framework used by https://github.com/DefinitelyTyped/DefinitelyTyped. The setup provided here is what is required by DefinitelyTyped. With that in mind, I would be more than happy to remove tests and additional deps and stick to only including the types themselves if this is preferable? Note also that the types include significant documentation for almost all methods and classes. This documentation _could_ be used to derive external documentation if that is desired.

Alternatively, if @jacobheun or any other maintainers would prefer to have these types with DefinitelyTyped, I'm happy to submit there instead!


types/index.d.ts
```diff
@@ -0,0 +1,324 @@
+/// <reference types="node" />
+
+/**
+ * Supported key types.
+ */
+export type KeyType = "Ed25519" | "RSA" | "secp256k1";
```
carsonfarmer: Note that, these are case-sensitive. They could be made case-insensitive by adding lower-/upper-case values. This goes for all `*Type` below.

types/index.d.ts
```diff
@@ -0,0 +1,324 @@
...
+   */
+  interface Digest {
+    digest(data: Buffer): Promise<Buffer>;
+    length: 20 | 32 | 64 | number;
+  }
+  /**
+   * Create a new HMAC Digest.
+   */
+  function create(
+    hash: "SHA1" | "SHA256" | "SHA512" | string,
```
carsonfarmer: Could be replaced with `HashType | string`...

types/index.d.ts
```diff
@@ -0,0 +1,324 @@
...
+   * Gets the ID of the key.
+   *
+   * The key id is the base58 encoding of the SHA-256 multihash of its public key.
+   * The public key is a protobuf encoding containing a type and the DER encoding
+   * of the PKCS SubjectPublicKeyInfo.
+   */
+  id(): Promise<string>;
+}
+
+export interface Keystretcher {
```
carsonfarmer: These three interfaces were 'created' for these types. They reflect the underlying structure of the return types. Their names could certainly be changed, and documentation could be added.

types/index.d.ts
```diff
@@ -0,0 +1,324 @@
...
+ * Currently the 'RSA' and 'ed25519' types are supported, although ed25519 keys
+ * support only signing and verification of messages. For encryption / decryption
+ * support, RSA keys should be used.
+ * Installing the libp2p-crypto-secp256k1 module adds support for the 'secp256k1'
+ * type, which supports ECDSA signatures using the secp256k1 elliptic curve
+ * popularized by Bitcoin. This module is not installed by default, and should be
+ * explicitly depended on if your project requires secp256k1 support.
+ */
+export namespace keys {
+  export {};
```
carsonfarmer: This is so only _certain_ interfaces etc are exported. Only those explicitly export will be visible to outside callers.

carsonfarmer: But this only applies at this top-level namespace, the lower level namespaces will have all their elements exported by default.

types/index.d.ts
```diff
@@ -0,0 +1,324 @@
...
+        equals(key: PrivateKey): boolean;
+        hash(): Promise<Buffer>;
+        id(): Promise<string>;
+        /**
+         * Exports the key into a password protected PEM format
+         *
+         * @param password The password to read the encrypted PEM
+         * @param format Defaults to 'pkcs-8'.
+         */
+        export(password: string, format?: "pkcs-8"): any;
```
carsonfarmer: Note: Export and Import are reserved words in ES6+, so they probably shouldn't be used in the future...

types/index.d.ts
```diff
@@ -0,0 +1,324 @@
...
+
+      function unmarshalEd25519PrivateKey(buf: Buffer): Promise<Ed25519PrivateKey>;
+      function unmarshalEd25519PublicKey(buf: Buffer): Ed25519PublicKey;
+      function generateKeyPair(): Promise<Ed25519PrivateKey>;
+      function generateKeyPairFromSeed(
+        seed: Buffer
+      ): Promise<Ed25519PrivateKey>;
+    }
+
+    namespace secp256k1 {
```
carsonfarmer: This class is only available if secp256k1 package is installed... but wasn't sure how to set this up in the types...

types/index.d.ts
```diff
@@ -0,0 +1,324 @@
...
+        id(): Promise<string>;
+      }
+
+      function unmarshalSecp256k1PrivateKey(bytes: Buffer): Promise<Secp256k1PrivateKey>;
+      function unmarshalSecp256k1PublicKey(bytes: Buffer): Secp256k1PublicKey;
+      function generateKeyPair(): Promise<Secp256k1PrivateKey>;
+    }
+  }
+
+  export const keysPBM: any;
```
carsonfarmer: This was left as `any` because it is quite a heavy definition, and seemed like it doesn't get used by external callers very much? If this is critical, it could be added..?

carsonfarmer: Or even as a separate PR as needed? 😉

types/index.d.ts
```diff
@@ -0,0 +1,324 @@
...
+   */
+  export function marshalPrivateKey(key: PrivateKey, type?: KeyType): Buffer;
+
+  /**
+   * Converts a PEM password protected private key into its representative object.
+   * @param pem Password protected private key in PEM format.
+   * @param password The password used to protect the key.
+   */
+  function _import(pem: string, password: string): Promise<supportedKeys.rsa.RsaPrivateKey>;
+  export { _import as import };
```
carsonfarmer: This little 'hack' is needed because import is a reversed word in TS.

carsonfarmer: Note that externally, the function is exported as `import`, **not** `_import`...

types/tsconfig.json
```diff
@@ -0,0 +1,27 @@
+{
```
carsonfarmer: This file is not _required_ unless we want linting capabilities _within_ this repo.

types/tslint.json
```diff
@@ -0,0 +1 @@
+{ "extends": "dtslint/dtslint.json" }
```
carsonfarmer: This file is not _required_ unless we want linting capabilities _within_ this repo.

types/types.spec.ts
```diff
@@ -0,0 +1,297 @@
+import crypto, {
```
carsonfarmer: This file is not _required_ unless we want linting/testing capabilities _within_ this repo. It was copied from the various *.spec.js` files within the repo. It is useful for making sure that the types match the usage of the module. However, it isn't ideal to have two versions of the 'tests' like this... At the moment, there isn't much of a nicer way to handle this that I'm aware of?

jacobheun:
> The setup provided here is what is required by DefinitelyTyped. With that in mind, I would be more than happy to remove tests and additional deps and stick to only including the types themselves if this is preferable?

Having the tests here makes sense to me. I was thinking it might be good to have a CI step to run these, but allow the failure for better visibility. The core libp2p team won't have bandwidth to maintain these, but if we can improve visibility into any breaking changes for the bindings we can proactively ping binding contributors to make them aware of the needed changes. 

@carsonfarmer do you think a CI step would be helpful for visibility? Will the linting catch breaking changes when the js api changes?

carsonfarmer:
Your question about testing and CI inspired me to try something out. And it is _much improved! Essentially now, this automatically uses the _existing_ tests to perform the type checking. And right now, it passes 'as is', which is really cool. That means the types match the way they are used _now_. This means, as code is changed, if the types no longer reflect the usage, it will get flagged if one runs `npm run test:types`. This could be turned on in CI (I haven't yet, didn't want to presume).

This is actually really really cool, and I haven't seen this done to this good effect elsewhere. It might provide a really nice template for other libp2p and IPFS projects looking to add types?

carsonfarmer:
So to answer your questions: 
> @carsonfarmer do you think a CI step would be helpful for visibility?
Yes!
> Will the linting catch breaking changes when the js api changes?
Yes!

test/keys/rsa.spec.js
```diff
@@ -88,15 +94,19 @@ describe('RSA', function () {
...
 
     const msg = Buffer.from('hello')
 
     // browser
-    const dec1 = await id.decrypt(Buffer.from('YRFUDx8UjbWSfDS84cDA4WowaaOmd1qFNAv5QutodCKYb9uPtU/tDiAvJzOGu5DCJRo2J0l/35P2weiB4/C2Cb1aZgXKMx/QQC+2jSJiymhqcZaYerjTvkCFwkjCaqthoVo/YXxsaFZ1q7bdTZUDH1TaJR7hWfSyzyPcA8c0w43MIsw16pY8ZaPSclvnCwhoTg1JGjMk6te3we7+wR8QU7VrPhs54mZWxrpu3NQ8xZ6xQqIedsEiNhBUccrCSzYghgsP0Ae/8iKyGyl3U6IegsJNn8jcocvzOJrmU03rgIFPjvuBdaqB38xDSTjbA123KadB28jNoSZh18q/yH3ZIg==', 'base64'))
```
carsonfarmer: Note that the type-checking helped to find this erroneous await :)

test/keys/secp256k1.spec.js
```diff
@@ -12,7 +12,8 @@ const crypto = require('../../src')
 
 describe('without libp2p-crypto-secp256k1 module present', () => {
   before(() => {
-    sinon.replace(crypto.keys.supportedKeys, 'secp256k1', null)
+    const empty = null
```
carsonfarmer: Needed for type checking... doesn't affect tests

carsonfarmer:
If you needed another reason to support types:
```
npx typedoc --mode file --includeDeclarations --excludeExternals ./src/index.d.ts
```
And feast your eyes on lovely docs!

carsonfarmer:
Yeh I'd love to get the types in ASAP as I'll use them right away, and they provide the basis for other types such as `peer-id` and a lot of others. The CI change is minimal, but would be good to do that in a separate PR so that we can evaluate its usefulness outside of the types themselves?

jacobheun:
Sounds good, released in **0.17.2**

carsonfarmer:
Woohoo! Cheers 👍

PR #166, opened 2020-02-14
perf: remove asn1.js and use node-forge

alanshaw:
We are already using `asn1` from `node-forge` in the code base. This PR removes the `asn1.js` dependency and uses the code we already imported from `node-forge` instead.

It should remove ~20 KB (gzipped) from the bundle.

Screenshot 2020-02-14 at 15 34 06

---

https://tools.ietf.org/html/rfc7518

### DIFF
--- a/src/keys/ecdh-browser.js
+++ b/src/keys/ecdh-browser.js
@@ -1,9 +1,9 @@
 'use strict'
 
 const errcode = require('err-code')
+const { Buffer } = require('buffer')
 const webcrypto = require('../webcrypto')
-const BN = require('asn1.js').bignum
-const { toBase64, toBn } = require('../util')
+const { bufferToBase64url } = require('../util')
 const validateCurveType = require('./validate-curve-type')
 
 const bits = {
@@ -81,6 +81,13 @@ const curveLengths = {
   'P-521': 66
 }
 
+const base64urlToBuffer = str => {
+  str = (str + '==='.slice((str.length + 3) % 4))
+    .replace(/-/g, '+')
+    .replace(/_/g, '/')
+  return Buffer.from(str, 'base64')
+}
+
 // Marshal converts a jwk encodec ECDH public key into the
 // form specified in section 4.3.6 of ANSI X9.62. (This is the format
 // go-ipfs uses)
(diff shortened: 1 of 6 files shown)


vasco-santos:
This seems to be failing on chrome 😢 I re-run both chrome jobs, but no luck! We need to investigate, because this would be a nice gain

alanshaw:
@vasco-santos aye, is still WIP, I’ve not even removed everything I need to in order to remove asn1.js yet!

I just submitted the PR for visibility and intent.

alanshaw:
> PASS  ./dist/index.min.js: 119.41KB < maxSize 155KB (gzip) 

🎉

test/util.spec.js
```diff
@@ -10,7 +10,7 @@ chai.use(dirtyChai)
 const util = require('../src/util')
 const BN = require('bn.js')
 
-describe('Util', () => {
+describe.skip('Util', () => {
```
alanshaw: Still TODO

src/keys/ecdh-browser.js
```diff
@@ -89,8 +96,11 @@ function marshalPublicKey (jwk) {
 
   return Buffer.concat([
     Buffer.from([4]), // uncompressed point
-    toBn(jwk.x).toArrayLike(Buffer, 'be', byteLen),
-    toBn(jwk.y).toArrayLike(Buffer, 'be', byteLen)
+    // TODO: BN.toArrayLike(Buffer, 'be', byteLen) restricts size to passed byteLen
```
alanshaw: Still TODO

src/util.js
```diff
@@ -1,20 +1,32 @@
...
+const forge = require('node-forge/lib/forge')
 
-// Convert a BN.js instance to a base64 encoded string without padding
-// Adapted from https://tools.ietf.org/html/draft-ietf-jose-json-web-signature-41#appendix-C
-exports.toBase64 = function toBase64 (bn, len) {
-  // if len is defined then the bytes are leading-0 padded to the length
-  const s = bn.toArrayLike(Buffer, 'be', len).toString('base64')
+exports.bigIntegerToBase64url = num => {
+  let byteArray = num.toByteArray()
+  // jsbn adds a 0 prefix for positive numbers, webcrypto does not like this
```
alanshaw: TODO verify

alanshaw:
Should be ready for review now

alanshaw:
Looks like this is also working upstream at https://github.com/ipfs/js-ipfs/pull/2785

PR #173, opened 2020-04-06
Integrate libp2p-crypto-secp256k1

wemeetagain:
Initially, [`libp2p-crypto-secp256k1`](https://github.com/libp2p/js-libp2p-crypto-secp256k1) was developed as an optional dependency.
As of c73adb00cc, `libp2p-crypto-secp256k1` is now a required dependency.

It now no longer makes sense to maintain `libp2p-crypto-secp256k1` as its own package, it makes sense to co-locate the code here, along with the implementations for rsa and ed25519 keys.

The intention is that `libp2p-crypto-secp256k1` can be retired/archived as a separate repo/package, and the functionality can be entirely here.

The commit history of `libp2p-crypto-secp256k1` has been maintained for posterity.


jacobheun:
libp2p and interop tests are all passing 🚀

jacobheun:
libp2p-crypto-secp256k1 has been archived and deprecation notices have been added to the repo and npm.

PR #176, opened 2020-06-09
chore: update node forge version

achingbrain:
I'm seeing two copies of `node-forge` in the js-IPFS bundle, my guess is because libp2p uses a `^` character for version selection and this module uses `~`.

The change here brings the version number into line with libp2p.

image


jacobheun:
Released in v0.17.7

Issue #145, opened 2019-03-14
Support `ed25519` and `secp256k1` in the import method


satazor:
Hello!

I was trying to publish a IPNS record using a `ed25519` key after seeing that support for `ed25519` and `secp256k1` keys was added here. I know that peer-id has to be changed to account for different key types, but I still decided to give it a try.

After trying, I concluded that `js-ipfs` calls [`crypto.keys.import`](https://github.com/ipfs/js-ipfs/blob/d2726fbca9da8ed139b75de0c6c0756bb4240109/src/core/components/name.js#L27) and fails because the import method assumes only RSA keys.

I can make a PR that adds support for `ed25519` and `secp256k1 ` keys by first extracting the private key info stored in the pem and react accordingly.

satazor:
//cc @vasco-santos @dignifiedquire @AlbertoElias

AlbertoElias:
That would be amazing if you could get that in. I didn't know that `import` would be necessary for IPNS, and as it wasn't supported for `ed25519` and `secp256k1` beforehand, I didn't worry about it in my PRs

satazor:
After checking `node-forge`, which we are using to import pkcs8 PEM files, I concluded it doesn't yet support them: https://github.com/digitalbazaar/forge/issues/667

I also discovered that they only support PKCS1 pem files and not PKCS8 pem files, meaning `libp2p-crypto` also only supports PKCS1, see https://github.com/digitalbazaar/forge/blob/5478021a7bdbbbc3b2ea981f28a0487b204cea10/lib/rsa.js#L1335. What they return there is a RSAPrivateKey ASN1 structure and then encode it in DER. This format is actually PKCS1, see https://tools.ietf.org/html/rfc3447#appendix-A.1.2. In contrast, PKCS9 evolved to support other key types by creating a wrapper which specified the key algorithm via the PrivateKeyInfo ASN1 structure, see https://tools.ietf.org/html/rfc5208#section-5.

To do this properly, we would need to support PKCS8 either by implementing in `node-forge` or in a library. I' more inclined in doing this in a library, which I actually already started doing. I will share some more details soon.

satazor:
The library is almost finished, here's a glimpse of it: https://github.com/ipfs-shipyard/js-crypto-key-composer/tree/initial-impl

[...]

PR #179, opened 2020-07-31
feat: add exporting/importing of non rsa keys in libp2p-key format

jacobheun:
This PR adds support for exporting/importing all keys in the `libp2p-key` format discussed in https://github.com/libp2p/js-libp2p-crypto/issues/145#issuecomment-661251436. This is designed to be a non breaking change. A future breaking change should update the default export of rsa to also be `libp2p-key` and provide an example for how to perform the migration of previously used keys.

Exported keys are encrypted via the AES-GCM algorithm using password derived PBKDF2 keys. They are exported as base64 encoded strings. I originally considered prefixing the exported bytes with a multibase prefix/multicodec prefix to make it easier to determine key type, but this would violate other formats like PEM, so we wouldn't get any meaningful reusability out of this.

This leverages Node.js and Webcrypto support only.

js-libp2p (including interop with go) and js-ipfs test suites are passing using this module. There pertinent PRs are referenced in the history below.

### DIFF
--- a/src/keys/rsa-class.js
+++ b/src/keys/rsa-class.js
@@ -111,11 +111,9 @@ class RsaPrivateKey {
    * Exports the key into a password protected PEM format
    *
    * @param {string} password - The password to read the encrypted PEM
-   * @param {string} [format] - Defaults to 'pkcs-8'.
+   * @param {string} [format=pkcs-8] - The format in which to export as
    */
   async export (password, format = 'pkcs-8') { // eslint-disable-line require-await
-    let key = null
-
     if (format === 'pkcs-8') {
       const buffer = new forge.util.ByteBuffer(this.marshal())
       const asn1 = forge.asn1.fromDer(buffer)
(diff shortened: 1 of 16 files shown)


jacobheun:
FYI, I am planning to release this as 0.17.9, if there are concerns let me know.

vasco-santos:
> A future non breaking change should update the default export of rsa to also be libp2p-key and provide an example for how to perform the migration of previously used keys.

You mean a future breaking change, right?

src/keys/rsa-class.js
```diff
@@ -112,25 +114,27 @@ class RsaPrivateKey {
...
+      const privateKey = forge.pki.privateKeyFromAsn1(asn1)
+
       const options = {
         algorithm: 'aes256',
         count: 10000,
         saltSize: 128 / 8,
         prfAlgorithm: 'sha512'
       }
-      pem = forge.pki.encryptRsaPrivateKey(privateKey, password, options)
+      key = forge.pki.encryptRsaPrivateKey(privateKey, password, options)
```
vasco-santos, line 130: Should we return here instead and not have the `let key`

src/keys/secp256k1-class.js
```diff
@@ -84,6 +87,21 @@ module.exports = (keysProtobuf, randomBytes, crypto) => {
       const hash = await this.public.hash()
       return multibase.encode('base58btc', hash).toString().slice(1)
     }
+
+    /**
+     * Exports the key into a password protected `format`
+     *
+     * @param {string} password - The password to encrypt the key
+     * @param {string} [format] - Defaults to 'libp2p-key'.
```
vasco-santos, line 95: Can we put the default as jsdoc?

PR #181, opened 2020-08-07
fix: remove rendundant public key

jacobheun:
BREAKING CHANGE: The private ed25519 key will no longer include the redundant public key

This is the breaking change portion of https://github.com/libp2p/js-libp2p-crypto/pull/178, which completes the coversion.


jacobheun:
@achingbrain since we need to do a breaking change release, I'd like to include this.

PR #183, opened 2020-12-15
chore: remove unused type

wemeetagain:
tested locally, I couldn't find any use of the node types here.

With this node type, there is a required peer dependency of `@types/node`


jacobheun (approved):
lgtm, this may have been previously needed with Buffer being a return type in a few places.

PR #185, opened 2021-03-15
Fix Ed25519 key ID generation (fixes ipfs/js-ipfs#3591)

nadimkobeissi:
Hello,

This commit proposes what appears to be a working and sane fix for the issue documented at ipfs/js-ipfs#3591.

I'd appreciate expert review and subsequent integration into the codebase.

PR #186, opened 2021-03-16
Fix Ed25519 PeerID generation

nadimkobeissi:
This commit pushes further fixes to the generation of Ed25519 peer IDs,
building upon the discussion in ipfs/js-ipfs#3591 and the subsequent
pull request libp2p/js-libp2p-crypto#185.

The purpose of this new pull request is to harmonize the encoding of
PeerIDs for Ed25519 keys such that the same new format is used
everywhere: peer IDs when assigned upon key generation, peer IDs when
shown via key listing, as well as the peer IDs displayed as IPNS names
when the key is used as the basis for an IPNS record.

Concretely, this changes the peer ID representation of Ed25519 keys from
the `Qm...` format to the newer `1...` format.

The accompanying test has been modified accordingly.

### DIFF
--- a/src/keys/ed25519-class.js
+++ b/src/keys/ed25519-class.js
@@ -4,8 +4,7 @@ const sha = require('multihashing-async/src/sha')
 const protobuf = require('protons')
 const errcode = require('err-code')
 const uint8ArrayEquals = require('uint8arrays/equals')
-const uint8ArrayToString = require('uint8arrays/to-string')
-
+const mh = require('multihashes')
 const crypto = require('./ed25519')
 const pbm = protobuf(require('./keys.proto'))
 const exporter = require('./exporter')
@@ -84,8 +83,8 @@ class Ed25519PrivateKey {
    * @returns {Promise<String>}
    */
   async id () {
-    const hash = await this.public.hash()
-    return uint8ArrayToString(hash, 'base58btc')
+    const encoding = mh.encode(this.public.bytes, 'identity')
+    return await mh.toB58String(encoding)
   }
 
   /**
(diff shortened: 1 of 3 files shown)


src/keys/ed25519-class.js
```diff
@@ -84,8 +83,8 @@ class Ed25519PrivateKey {
    * @returns {Promise<String>}
    */
   async id () {
-    const hash = await this.public.hash()
```
aschmahmann, line 87: Should the `hash` function on the public key return the identity hash? The public and private keys still have sha2-256 hash functions associated with them.

cc @jacobheun @vasco-santos

nadimkobeissi: @aschmahmann For what it's worth, the peer ID generated here matches those generated for signed IPNS records, and so this appears to follow the correct spec.

aschmahmann: Sure, you've made the `id` function behave correctly. However, I'm not sure what the `public.hash()` function is used for and that is potentially another bug.

nadimkobeissi: `public.hash()` calls `sha.multihashing(this.bytes, 'sha2-256')`, effectively returning a SHA2-based multihash of the public key. I used this call specifically because it followed the spec, and because the same call was used elsewhere in functions of the same depth in the same prototype.

jacobheun: The hash should likely just be returning the identity hash, but that would be breaking across the api so that should happen outside of this, perhaps when we revisit reducing the bundle size of this repo for browsers.

This change looks good and fixes the existing bug.

jacobheun (approved):
LGTM, thanks for patching!

PR #187, opened 2021-04-07
chore: update deps

vasco-santos:
Updates dependencies to the latest versions to dedupe the dependency tree.

This also removes protons in favour of protobufjs for smaller bundles


rvagg (approved):
👌 seems pretty straightforward aside from the protobufjs change

achingbrain (approved):

Issue #190, opened 2021-06-29
Proposal: Separate key export functionality


Gozala:
I've been working on updating this implementation to align it with https://github.com/libp2p/js-libp2p-interfaces/pull/100 and I am noticing that `export` function on private keys introduces unnecessary dependencies as in export functionality is is not key specific. So I would suggest factoring that functionality out into separate interface e.g.

```ts
interface KeyExporter<Format extends string = string> {
   format: Format
   export(key:PrivateKey, password?:Uint8Array): Uint8Array
}
```

This would make remove indirection and instead of doing e.g. 

```js
key.export('my secret', 'pkcs-8')
key.export('my secret', 'libp2p-key')
```

One would be a able to do following instead:

```js
pkcs8.export(key, 'my secret')
libp2pExporter.export(key, 'my secret')
```

We did a similar change in js-multiformats for base encoding and hashing where instead of passing identifiers you pass references and it simplified things greatly.

Gozala:
Also if we really want to we could still keep `export` method on PrivateKey but we could just swap format identifier with `KeyExporter` as described above so that same example could be rewritten as follows:

```js
key.export('my secret', pkcs8)
key.export('my secret', libp2pKey)
```

Gozala:
I see now that `pkcs8` uses rsa specific node-forge calls

https://github.com/libp2p/js-libp2p-crypto/blob/117b52040a384e6661283c97dbe6cff0d136c686/src/keys/rsa-class.js#L120-L128

[...]

PR #191, opened 2021-06-29
feat: type with libp2p-interfaces

Gozala:
This is fallout of the https://github.com/libp2p/js-libp2p/issues/955 effort. More specifically it types everything via crypto interfaces definitions from https://github.com/libp2p/js-libp2p-interfaces/pull/100 which is necode to eded to align it with a pull that integrates interfaces in js-libp2p

# Reviewer notes

- This removes hand written typedefs and generates those instead with aegir (so everything is typed now)
- Bunch of places did something like `new Uint8Array(buffer, buffer.byteOffset, buffer.byteLength)` where `buffer` was instance of `ArrayBuffer`which does not have `byteOffset` property. It worked correctly because `undefined` was triggering same code path as `new Uint8Array(buffer)`
- Private/Public keys were added `algorithm` property (introduced by interface) which follows the same pattern as base encoders, hashers and block codecs (in multiformats). This allows narrow down type without resorting to `instanceof`, which is not a thing when embracing interfaces as opposed to implementations.
- Some functions took e.g. `CurveType | string` style args, now they are strict and just take `CurveType` which enables type checker to catch problems at compile time. Users generally can cast string to specific literals when needed.
- `src/keys/keys.d.ts` is generated now via `pbjs`
- Bunch of ts-ignores are used when importing node-forge sub-modules because there are not typedefs for those. I assume reason why we don't simply `require('node-forge')` is so that we can avoid pulling unnecessary code, if that is not the case maybe we should switch to using it instead.
- Had to hoist `Secp256k1PublicKey` / `Secp256k1PrivateKey` because TS doesn't support non static (generate) classes as their type can't really be expressed. I would personally not even expose classes which would also address the issue, but that seemed to invasive change.
  - Do we even need the whole factory thing there, seems like it unnecessarily complication.
- `generateKeyPairFromSeed` no longer takes 3rd parameter because nothing was using it.
- RSA keys had strange `genSecret` method. It made little sense to me so I've omitted it from the interface (no other key implements it) and created a static function on the module which is what test is using now. I kept the method around just in case, but type checker will complain if you try to use it.
- I've marked some functions (that made little sense to me) deprecated, they will just show-up crossed out in the vscode. See more in suggestions

# Suggestions

- I feel like `src/keys/index.js` attempts to provide a generic API where it makes no sense. E.g. `generateKeyPairFromSeed` can only be used by `Ed25519` keys, so it throws in all other cases. `generateKeyPair` ignores bits when `secp256k1` is used.

   All this is just suggests that those APIs aren't generic, it would make a lot more sense to just use `ed25519.generateKeyPairFromSeed` and not have generic equivalent. I think same is true for `generateKeyPair` there is questionable value in indirection here:

   ```js
   keys.generateKeyPairFromSeed('Ed25519', seed)
   // VS
   ed25519.generateKeyPairFromSeed(seed)
   ```

  Which is also true for `generateKeyPair`.
- I think export function on private keys is another source of indirection that complicates things instead of simplifying them more on this in #190


src/keys/key-stretcher.js
```diff
@@ -34,6 +43,7 @@ module.exports = async (cipherType, hash, secret) => {
     throw errcode(new Error('missing hash type'), 'ERR_MISSING_HASH_TYPE')
   }
 
+  // @ts-expect-error - Blowfish has no keySize
```
Gozala, line 46: @vasco-santos this seems like a bug that needs addressing

src/keys/rsa-class.js
```diff
@@ -115,17 +171,17 @@ class RsaPrivateKey {
    */
   async export (password, format = 'pkcs-8') { // eslint-disable-line require-await
     if (format === 'pkcs-8') {
+      // @ts-ignore - ByteBuffer isn't present in typedefs
       const buffer = new forge.util.ByteBuffer(this.marshal())
       const asn1 = forge.asn1.fromDer(buffer)
       const privateKey = forge.pki.privateKeyFromAsn1(asn1)
 
-      const options = {
+      return forge.pki.encryptRsaPrivateKey(privateKey, password, {
```
Gozala, line 179: Note: This allows TS to infer literals instead of treating them as strings.

BigLep:
2021-11-12 update: we do still want this work, but it needs to be looked at in a couple weeks as part of larger js-libp2p typing changes that will be occurring.

PR #192, opened 2021-07-06
chore: upgrades to new multiformats module

achingbrain:
Removes multibase/multicodec/multihashing in favour of the implementations in the new multiformats module.

Turns out all we really use is sha2-256 and identity, we don't need all the other stuff it pulls in.


vasco-santos (approved):
LGTM

PR #202, opened 2021-10-13
feat: use noble-secp256k1 and noble-ed25519

hugomrdias:
https://github.com/paulmillr/noble-secp256k1#noble-secp256k1--
https://github.com/paulmillr/noble-ed25519

116KB -> 70.32KB


hugomrdias:
@achingbrain @vasco-santos should be good to go

src/keys/rsa-class.js
```diff
@@ -6,7 +6,7 @@ const { equals: uint8ArrayEquals } = require('uint8arrays/equals')
 const { toString: uint8ArrayToString } = require('uint8arrays/to-string')
 
 require('node-forge/lib/sha512')
-require('node-forge/lib/ed25519')
+// require('node-forge/lib/ed25519')
```
vasco-santos, line 9:
```suggestion
```

src/keys/ed25519.js
```diff
@@ -1,24 +1,59 @@
...
+  const privateKey = concatKeys(privateKeyRaw, publicKey)
+
+  return {
+    privateKey,
+    publicKey
+  }
+}
+
+exports.hashAndSign = function (privateKey, msg) { // eslint-disable-line require-await
+  const privateKeyRaw = privateKey.slice(0, 32)
```
vasco-santos, line 43: should we make above export for key length constants and replace 32 here and in concatKeys by it?

test/keys/secp256k1.spec.js
```diff
@@ -210,27 +217,19 @@ describe('crypto functions', () => {
...
     try {
+      // @ts-ignore
       await secp256k1Crypto.hashAndVerify(privKey, sig, null)
     } catch (err) {
       return // expected
     }
     throw new Error('Expected error to be thrown')
   })
 
-  it('errors when validating a message with an invalid signature', async () => {
```
vasco-santos, line 220: why removing this?

vasco-santos: Oh I see, so this now returns a boolean `false` instead of throwing an error. Can we flag that breaking change in the commit message?

PR #211, opened 2021-11-27
feat: use web crypto for Ed25519 keys in node.js if available

achingbrain:
Node.js 15 added support for Ed25519 keys in web crypto so let's use that instead of pure js implementations which are slow.

I've also added a benchmark suite so we can compare implementations:

```console
$ node ed25519.js
native ed25519 x 4,584 ops/sec ±2.09% (79 runs sampled)
noble ed25519 x 193 ops/sec ±33.95% (79 runs sampled)
node-forge ed25519 x 60.75 ops/sec ±3.36% (68 runs sampled)
node.js ed25519 x 3,320 ops/sec ±2.33% (78 runs sampled)
```

Note the [ed25519 module](https://www.npmjs.com/package/ed25519) actually comes out on top but it's API is synchronous - this is probably why it's faster but will likely impede performance elsewhere.

Falls back to the pure-js implementation when Ed25519 is not available (e.g. node < 15, electron, etc).


dapplion:
d577htl-144b9110-7f54-418b-a656-7e72644873b7

achingbrain:
Added a few more competing implementations to the benchmarks.  The wasm version is quite the eye-opener:

```console
$ node ed25519.js
@noble/ed25519 x 125 ops/sec ±32.00% (55 runs sampled)
@stablelib/ed25519 x 48.78 ops/sec ±8.33% (63 runs sampled)
node-forge/ed25519 x 54.44 ops/sec ±5.54% (69 runs sampled)
supercop.wasm x 3,651 ops/sec ±1.33% (83 runs sampled)
ed25519 (native module) x 4,813 ops/sec ±0.86% (86 runs sampled)
node.js web-crypto x 3,466 ops/sec ±2.04% (80 runs sampled)
```

Issue #205, opened 2021-10-23
Use noble-hashes for pbkdf2, sha, hmac


paulmillr:
Hey, saw #202. I've also made [noble-hashes](/paulmillr/noble-hashes) recently, which implements all primitives in very minimal format.

It's going to be audited soon, unlike `forge`, so you should switch to it.

paulmillr:
It's been audited now, see [the blog post](https://medium.com/nomic-labs-blog/a-safer-smaller-and-faster-ethereum-cryptography-stack-5eeb47f62d79).

PR #218, opened 2021-11-30
feat: replace keypair and ursa-optional with web-crypto for node

achingbrain:
Removes native module `ursa-optional` and js fallback library `keypair` in favour of using web-crypto which is available in node 15+

Also adds benchmark suite similar to Ed25519:

```console
$ node index.js
ursa-optional x 323 ops/sec ±2.84% (70 runs sampled)
keypair x 55.33 ops/sec ±11.75% (59 runs sampled)
node.js web-crypto x 300 ops/sec ±2.29% (74 runs sampled)
fastest is ursa-optional
```

BREAKING CHANGE: requires node 15+


hugomrdias:
We have an issue with webcrypto not supporting our RSA algo for encryption, i will find some refs on this tomorrow

hugomrdias:
https://github.com/libp2p/js-libp2p-crypto/issues/205#issuecomment-950294099

Found it

achingbrain:
That's interesting, so in the browser we fall back to node-forge for RSA encryption/decryption because it supports PKCS1 padding, which we've [hard coded](https://github.com/libp2p/js-libp2p-crypto/blob/42ed07a82a58ffc26be2667ec6e24ac8604b49cf/src/keys/rsa.js#L75) in the node.js version.  As it stands this PR will make node do the same thing which will be slower than using `ursa-optional`.

Perhaps we could use node crypto for encrypt/decrypt to retain the performance but keypair generation and sign/verify with web crypto?  They we can still get rid of the extra deps.

achingbrain:
Or just `crypto.generateKeyPair`?

achingbrain:
Superseded by #219

PR #219, opened 2021-12-01
feat: replace keypair and ursa-optional with node crypto

achingbrain:
We use `ursa-optional` to geenrate RSA keypairs in node.js, falling back to `keypair` if compilation of `ursa-optional` failed at install time.

Node.js has supported generating RSA keypairs since 10.x so use that instead and remove the `ursa-optional` and `kepair` dependencies.

As a bonus node 15+ also supports `jwk` format keys everywhere so the `pem-jwk` dep can go too.

BREAKING CHANGE: requires node 15+

### DIFF
--- a/src/keys/rsa.js
+++ b/src/keys/rsa.js
@@ -1,43 +1,23 @@
 'use strict'
 
 const crypto = require('crypto')
+const promisify = require('util').promisify
 const errcode = require('err-code')
 const randomBytes = require('../random-bytes')
-// @ts-check
-/**
- * @type {PrivateKey}
- */
-let keypair
-try {
-  if (process.env.LP2P_FORCE_CRYPTO_LIB === 'keypair') {
-    throw new Error('Force keypair usage')
-  }
-
-  const ursa = require('ursa-optional') // throws if not compiled
-  keypair = ({ bits }) => {
-    const key = ursa.generatePrivateKey(bits)
-    return {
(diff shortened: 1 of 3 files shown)


src/keys/rsa.js
```diff
@@ -1,43 +1,23 @@
...
-}
-const pemToJwk = require('pem-jwk').pem2jwk
-const jwkToPem = require('pem-jwk').jwk2pem
+const keypair = promisify(crypto.generateKeyPair)
 
 exports.utils = require('./rsa-utils')
 
 exports.generateKey = async function (bits) { // eslint-disable-line require-await
-  const key = keypair({ bits })
+  const key = await keypair('rsa', {
```
achingbrain, line 12: Could use the sync version here, but used the async version seeing as how the function is async anyway.

achingbrain:
Can put `pem-jwk` back in if we care about node 14 compatibility, but we shouldn't as 16 is now LTS.

hugomrdias (approved):
LGTM 🔥🔥

hugomrdias:
Now we just need to find a way to work around the RSA encryption thing to take out forge 😭😭

vasco-santos (approved):

achingbrain:
@hugomrdias some thoughts on that here: https://github.com/libp2p/js-libp2p-crypto/issues/220

PR #230, opened 2022-01-06
chore: change min node support to 16

### LINKED ISSUE #221: Node 15: The property 'options.publicKeyEncoding.format' is invalid. Received 'jwk'
Latest release says that only node 15+ is supported: https://github.com/libp2p/js-libp2p-crypto/releases/tag/v0.21.0

Note that the `package.json` was not updated accordingly:

https://github.com/libp2p/js-libp2p-crypto/blob/508c790c52f15d8194f926b0d56c0069caf4ab40/package.json#L65

Node 15 is actually not supported as we can see when running the CI with 15: https://github.com/D4nte/js-libp2p-crypto/runs/4438572378?check_suite_focus=true

```
TypeError [ERR_INVALID_ARG_VALUE]: The property 'options.publicKeyEncoding.format' is invalid. Received 'jwk'
```

Only node 16 works.

I wonder if it's due to this bug only fixed on node 16: https://github.com/nodejs/node/issues/39205

D4nte:
This library does not work with node 15 (#221).
Moreover, node 15 has reached End-of-Life.

Resolves #221.

### DIFF
--- a/package.json
+++ b/package.json
@@ -87,7 +87,7 @@
     "wherearewe": "^1.0.0"
   },
   "engines": {
-    "node": ">=15.0.0"
+    "node": ">=16.0.0"
   },
   "repository": {
     "type": "git",


achingbrain:
Weird that, `'jwk'` is a supported option for node 15 [as per the docs](https://nodejs.org/docs/latest-v15.x/api/crypto.html#crypto_keyobject_export_options).

Officially these modules support [LTS and Current](https://github.com/ipfs/community/blob/master/CONTRIBUTING_JS.md#supported-versions) only so this makes sense.

achingbrain (approved):

D4nte:
> Weird that, `'jwk'` is a supported option for node 15 [as per the docs](https://nodejs.org/docs/latest-v15.x/api/crypto.html#crypto_keyobject_export_options).
> 
> Officially these modules support [LTS and Current](https://github.com/ipfs/community/blob/master/CONTRIBUTING_JS.md?rgh-link-date=2022-01-08T06%3A58%3A59Z#supported-versions) only so this change makes sense.

Yes but it's a bug (https://github.com/nodejs/node/issues/39205) that prevents the usage of `'jwk'`. The bug was fixed for 16 after EOL of 15.

PR #239, opened 2022-01-18
fix: pin @noble/ed25519 version

achingbrain:
`@noble/ed25519@1.5.0` introduces a regression around signatures so
pin to `1.4.0` while it is resolved.

Refs: https://github.com/paulmillr/noble-ed25519/issues/45


paulmillr:
Can be reverted now

PR #245, opened 2022-02-15
fix: bump noble-ed25519 to an audited version

paulmillr:
This makes sure it's always 1.6.0+ and not just 1.4.0 if some dep is also using 1.4.0

https://github.com/paulmillr/noble-ed25519/releases/tag/1.6.0

### DIFF
--- a/package.json
+++ b/package.json
@@ -171,7 +171,7 @@
     "build:proto:types": "pbts -o src/keys/keys.d.ts src/keys/keys.js"
   },
   "dependencies": {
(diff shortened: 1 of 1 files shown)


hugomrdias (approved):

hugomrdias (approved):

achingbrain (approved):

paulmillr:
@achingbrain it would be great if 0.22.8 was released, and then peer-id 0.16.1 was released, because we depend on this dependency chain in @ethereumjs

PR #255, opened 2022-05-12
feat: use @noble/hashes instead of node-forge for keys

### LINKED ISSUE #205: Use noble-hashes for pbkdf2, sha, hmac

achingbrain:
Use @noble/hashes instead of node-forge for keys

Fixes #205

### DIFF
--- a/package.json
+++ b/package.json
@@ -171,6 +171,7 @@
   "dependencies": {
     "@libp2p/interfaces": "^1.3.20",
     "@noble/ed25519": "^1.6.0",
+    "@noble/hashes": "^1.0.0",
     "@noble/secp256k1": "^1.5.4",
     "err-code": "^3.0.1",
     "iso-random-stream": "^2.0.0",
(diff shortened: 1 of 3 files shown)


BigLep:
2022-09-13 conversation: we'd prioritize this if it would actually let us get rid of node-forge.  If we were to merge it now the bundle size would increase.

paulmillr:
@BigLep 

> it would actually let us get rid of node-forge

I have just checked and the PR actually throws node-forge out. It is not included.

Can you explain what you mean by bundle size increase?

ljharb:
@paulmillr in that one file, but not out of the entire package, i believe https://github.com/search?q=repo%3Alibp2p%2Fjs-libp2p-crypto%20node-forge&type=code

paulmillr:
@ljharb ah, thought the pr is complete. Thanks.

PR #262, opened 2022-07-20
fix: remove iso-random-stream dependency

D4nte:
This dependency brings in NodeJS's `Buffer` which then needs to be polyfilled in the browser.

`@noble/*` already brings a utility that switch between Node's and the browser's `randomBytes` interfaces.

One caveat is that `iso-random-stream` enables generation of arrays greater than 65536 in length in the browser.
It does not strike to me as a feature needed for libp2p-crypto. Let me know your thoughts.

### DIFF
--- a/benchmark/ed25519/compat.cjs
+++ b/benchmark/ed25519/compat.cjs
@@ -16,12 +16,12 @@ const forge = require('node-forge/lib/forge')
  * function because key generation is deterministic for a given seed.
  */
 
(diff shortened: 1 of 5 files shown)


wemeetagain (approved):
lgtm

hugomrdias (approved):

hugomrdias:
@achingbrain

achingbrain (approved):

PR #269, opened 2022-08-01
fix: allow ed25519 to be instantiated with non concatenated key

### LINKED ISSUE #200: Ed25519 Private Key length
Constructing a new Ed25519 key by passing a 32 byte Uint8Array should be allowed.  I assume this check is due to the convention of passing in private & public concatenated, would be nice to detect a 32 byte key [per the spec](https://datatracker.ietf.org/doc/html/rfc8032#section-5.1.5).

I could work on a pull if this is of interest to anyone else. 

https://github.com/libp2p/js-libp2p-crypto/blob/a99f419f7904317271c3d0eb6f8eb31e69d3f9b5/src/keys/ed25519-class.js#L44

mpetrunic:
resolves #200

### DIFF
--- a/src/keys/ed25519-class.ts
+++ b/src/keys/ed25519-class.ts
@@ -1,11 +1,11 @@
 import errcode from 'err-code'
-import { equals as uint8ArrayEquals } from 'uint8arrays/equals'
-import { sha256 } from 'multiformats/hashes/sha2'
 import { base58btc } from 'multiformats/bases/base58'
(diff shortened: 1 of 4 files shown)


wemeetagain (approved):

BigLep:
2022-09-13 triage conversation: before merging, manually sanity check running this change against the js-libp2p testsuite to avoid any surprises.

PR #278, opened 2022-10-05
fix: use a more intuitive err-code

tabcat:
Related: https://github.com/libp2p/js-libp2p/issues/1269

This PR was created in js-libp2p-crypto to dial-in/double-check what is needed before making more changes in other repos.

Uses a more intuitive [err-code](https://github.com/IndigoUnited/js-err-code); the repo is currently located at [coderra](https://github.com/tabcat/coderra).

The issue OP's implementation is used. The parameters and behavior have changed compared with err-code.

##### [src/index.ts](https://github.com/tabcat/coderra/blob/master/src/index.ts):

```js
// https://github.com/libp2p/js-libp2p/issues/1269#issue-1280763716
export default class CodeError<T extends Record<string, any>> extends Error {
  code: string
  props: T

  constructor (message: string, code: string, props?: T) {
    super(message)

    this.code = code
    this.props = props ?? {} as T // eslint-disable-line @typescript-eslint/consistent-type-assertions
  }
}
```

CodeError behavior vs. errcode:

- string required as first param, not an Error
- `message`: will not throw at runtime if given something other than a string
- `code`: will not throw at runtime if given something other than a string, required to be a string by typescript/cannot sometimes be `props`
- `props`: properties of props are not assigned to directly Error instance
- Cannot be used to throw TypeError's

### DIFF
--- a/package.json
+++ b/package.json
@@ -181,7 +181,6 @@
     "@libp2p/interface-keys": "^1.0.2",
     "@noble/ed25519": "^1.6.0",
     "@noble/secp256k1": "^1.5.4",
-    "err-code": "^3.0.1",
     "multiformats": "^9.4.5",
     "node-forge": "^1.1.0",
     "protons-runtime": "^3.1.0",
--- a/package.json
+++ b/package.json
@@ -181,6 +181,7 @@
     "@libp2p/interface-keys": "^1.0.2",
     "@noble/ed25519": "^1.6.0",
     "@noble/secp256k1": "^1.5.4",
+    "coderra": "github:tabcat/coderra",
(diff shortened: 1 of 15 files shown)


tabcat:
Going to make the repo look better and add more tests sometime in the next few days. will then mark undrafted.

tabcat:
Closing from lack of input. Also don't like the `coderra` name anymore, `error-code` as a package name for something like this seems much better.

wemeetagain:
This looks good. I'm curious about how the generic type works in practice. (This repo doesn't seem to use the "props")

Tag me when you open another PR and I'll be sure to give you a review.

PR #284, opened 2022-11-11
chore: replace err-code with error-code

tabcat:
Replaces https://github.com/libp2p/js-libp2p-crypto/pull/278 which goes into more detail

Tried to find a libp2p repo that uses the props feature but was unsuccessful. Added the generic usage for props to [code-err](https://github.com/tabcat/code-err) tests.

~~error-code package is not published yet but documentation has been added. Also the class is named and imported as `ErrorCode` instead of `CodeError` like the OP; can change it back but thought this worked fine.~~

@wemeetagain 

Before Merging:

- [x] https://github.com/tabcat/error-code/issues/1
- [x] publish and install `code-err` package from npm
- [ ] get feedback on changes made here before other libp2p packages
- [ ] create issue to track all related PRs for libp2p packages, including this one

### DIFF
--- a/package.json
+++ b/package.json
@@ -181,7 +181,7 @@
     "@libp2p/interface-keys": "^1.0.2",
     "@noble/ed25519": "^1.6.0",
     "@noble/secp256k1": "^1.5.4",
-    "err-code": "^3.0.1",
+    "code-err": "^1.0.0",
     "multiformats": "^10.0.0",
     "node-forge": "^1.1.0",
     "protons-runtime": "^4.0.1",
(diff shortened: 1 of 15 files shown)


tabcat:
How do these changes look? Anything needed before doing the same to the other libp2p packages?

tabcat:
Happy to install into libp2p* packages this week if everything is fine. Can talk about supply-chain/repo ownership if concerned. Email is on my profile.

wemeetagain:
Hey we just discussed this and the supply-chain issue in our triage call. We think it would be good to maintain this under the libp2p org.
Can you make a PR to add the `CodeError` code here? https://github.com/libp2p/js-libp2p-interfaces/blob/master/packages/interfaces/src/errors.ts
This would make `CodeError` available like so:
`import { CodeError } from '@libp2p/interfaces/errors'`

PR #313, opened 2023-03-24
fix: deriveKey in webkit linux (workaround)

MarcoPolo:
After some painful debugging (see: https://github.com/libp2p/js-libp2p/pull/1627#issuecomment-1483401928) I've found out why js-libp2p fails on webkit in Linux. The subtlecrypto implementation doesn't like to derive a key from an empty imported key. This works on all other browsers, so it seems like the webkit implementation is doing the wrong thing. Maybe worth opening a bug and writing a test for them. In the mean time here's a workaround.

This unblocks webkit testing in the interop tester (which runs on linux). This also lets folks use js-libp2p from webkit based linux browsers, although it doesn't seem like anyone else ran into this issue.

### DIFF
--- a/package.json
+++ b/package.json
@@ -172,6 +172,7 @@
     "test:chrome-webworker": "aegir test -t webworker",
     "test:firefox": "aegir test -t browser -- --browser firefox",
     "test:firefox-webworker": "aegir test -t webworker -- --browser firefox",
+    "test:webkit": "aegir test -t browser -- --browser webkit",
     "test:node": "aegir test -t node --cov",
     "test:electron-main": "aegir test -t electron-main",
     "release": "aegir release",


MarcoPolo:
Nice. This surfaced a proper error in macos webkit. Surprisingly, that test passes in linux webkit.

MarcoPolo:
Per sync discussion:

A better fix will be to use a static key that is generated from a working implementation when the password is "". When the password is "" we know what the derived password is, so we can fill it in with a constant. This works around the issue in webkit linux by not having it try to generate that constant which fill fail.

src/ciphers/aes-gcm.browser.ts
```diff
@@ -29,10 +45,15 @@ export function create (opts?: CreateOptions): AESCipher {
       password = fromString(password)
     }
 
+    let cryptoKey
```
achingbrain, line 48: Prevents accidental assignment of an unrelated type in the future.

```suggestion
    let cryptoKey: CryptoKey
```

achingbrain (approved):
LGTM, just a couple of nits for the types.

